PR #17, opened 2024-11-23
bug - resolve build issues with nextjs

setbern:
Fix build issues for nextjs

### DIFF
--- a/src/util/atoms.ts
+++ b/src/util/atoms.ts
@@ -7,8 +7,10 @@ import { StacksNetwork, STACKS_TESTNET } from "@stacks/network";
 
 export const store = createStore();
 
-const BITCOIND_URL =
+export const BITCOIND_URL =
   process.env.NEXT_PUBLIC_BITCOIND_URL || "http://localhost:18443";
+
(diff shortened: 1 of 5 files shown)


src/util/atoms.ts
```diff
@@ -2,12 +2,14 @@
...
 import { AppConfig, UserSession, UserData } from "@stacks/connect";
-import { BITCOIND_URL } from "@/app/api/proxy/[...proxy]/route";
 
 import { StacksNetwork, STACKS_TESTNET } from "@stacks/network";
 
 export const store = createStore();
 
+const BITCOIND_URL =
+  process.env.NEXT_PUBLIC_BITCOIND_URL || "http://localhost:18443";
+
```
aldur, line 12: nit: duplicated code can be easily removed away by keeping it in a common file and importing it

aldur (approved):

src/util/atoms.ts
aldur: Fixed in [a9695fd](https://github.com/stacks-network/sbtc-bridge/pull/17/commits/a9695fd8bb6c7eed9142330d3c98d2a1cbbd9894)

PR #25, opened 2024-11-25
Add Dockerfile and fix handling of environmental variables

aldur:


aldur:
Note: this is in draft because when run in build (as opposed to dev) mode, Next.JS embeds env variables into the code at build-time. But we want to set variables at run-time, so at build-time they are unset and the code doesn't work.

Ref: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables

aldur:
This has now been deployed, thanks!

hozzjss (approved):

PR #29, opened 2024-11-27
965 when the bridge is in testnet mode inform users on how to get regtest bitcoin

hozzjss:
As a user I want to get tBTC to test the bridge on the sbtc testnet env

resolves https://github.com/stacks-network/sbtc/issues/965

### DIFF
--- a/src/comps/Header.tsx
+++ b/src/comps/Header.tsx
@@ -19,7 +19,10 @@ import { useWallet } from "@/util/WalletContext";
 import { NotificationStatusType } from "./Notifications";
 import { GetTestnetBTC } from "./get-testnet-btc";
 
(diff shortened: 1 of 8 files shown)


src/comps/Header.tsx
```diff
@@ -17,6 +17,9 @@ import {
 import ConnectWallet from "./ConnectWallet";
 import { useWallet } from "@/util/WalletContext";
 import { NotificationStatusType } from "./Notifications";
+import { GetTestnetBTC } from "./get-testnet-btc";
+
+const isTestnet = process.env.NEXT_PUBLIC_WALLET_NETWORK !== "mainnet";
```
aldur, line 22: Since this is used to enable the regtest faucet button and that only works for sBTC testnet, I'd check that the variable is set to `sbtcTestnet` specifically.

hozzjss: fixed

PR #32, opened 2024-11-28
Allow the bridge to prevent the user from depositing if it would violate a cap

hozzjss:
## Description

As a user I want to submit a sBTC bridge request only if the deposit amount does not exceeds the limits for deposit. This is an important validation for bridging sBTC so that btc isn't lost or extra work is needed to retrieve it. This PR adds an action to get the limits from the emily api, and a client side validation for the btc amount input.

https://github.com/user-attachments/assets/30b2191b-7cb2-4f3e-b162-8f9c5428fdfd

image

For details refer to issue https://github.com/stacks-network/sbtc/issues/859

## Type of Change
- Other

No

## Are documentation updates required?
No

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,8 @@
     "@bitcoinerlab/secp256k1": "1.1.1",
     "@headlessui/react": "2.1.2",
     "@heroicons/react": "2.1.5",
-    "@stacks/transactions": "7.0.2",
+    "@stacks/transactions": "^7.0.2",
+    "@tanstack/react-query": "^5.61.5",
     "add": "2.0.6",
     "babel-loader": "9.2.1",
     "c32check": "2.0.0",
@@ -28,7 +29,8 @@
     "sats-connect": "3.0.1",
     "server-only": "^0.0.1",
     "webpack": "5.95.0",
-    "yarn": "1.22.22"
+    "yarn": "1.22.22",
+    "yup": "^1.4.0"
   },
   "devDependencies": {
     "@leather.io/rpc": "^2.1.20",
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "server-only": "^0.0.1",
     "webpack": "5.95.0",
     "yarn": "1.22.22",
-    "yup": "^1.4.0"
+    "yup": "1.4.0"
   },
   "devDependencies": {
     "@leather.io/rpc": "^2.1.20",
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "sats-connect": "3.0.1",
+    "server-only": "^0.0.1",
     "webpack": "5.95.0",
     "yarn": "1.22.22"
   },
--- a/src/actions/get-current-sbtc-supply.ts
+++ b/src/actions/get-current-sbtc-supply.ts
@@ -1,0 +1,24 @@
+"use server";
+
+import { env } from "@/env";
+import {
+  fetchCallReadOnlyFunction,
+  type ResponseOkCV,
+  type UIntCV,
+} from "@stacks/transactions";
+import { type StacksNetworkName } from "@stacks/network";
+export default async function getCurrentSbtcSupply() {
+  const response = (await fetchCallReadOnlyFunction({
+    contractAddress: env.SBTC_CONTRACT_DEPLOYER!,
+    contractName: "sbtc-token",
+    functionName: "get-total-supply",
+    functionArgs: [],
+    network:
+      env.WALLET_NETWORK === "sbtcDevenv"
+        ? "devnet"
+        : (env.WALLET_NETWORK as StacksNetworkName),
+    senderAddress: env.SBTC_CONTRACT_DEPLOYER!,
+  })) as ResponseOkCV<UIntCV>;
+
+  return response;
+}
--- a/src/actions/get-emily-limits.ts
+++ b/src/actions/get-emily-limits.ts
@@ -1,0 +1,23 @@
+"use server";
+
+import { env } from "@/env";
+type EmilyLimits = {
+  pegCap: null | number;
+  perDepositCap: null | number;
+  perWithdrawalCap: null | number;
+  accountCaps: AccountCaps;
+};
+
+type AccountCaps = {};
+
+export default async function getEmilyLimits() {
+  const res = await fetch(`${env.EMILY_URL}/limits`);
+  const json = (await res.json()) as EmilyLimits;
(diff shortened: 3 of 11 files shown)


aldur:
I think this is a good start, but:

1. If the limit is already reached, then the UI should display it clearly and disable all user actions.
2. It could also happen that the cap hasn't been reached when the user submits the deposit, but _then_ it will be reached. In that case, they need to be able to reclaim their Bitcoin.

hozzjss:
> I think this is a good start, but:
> 
> 1. If the limit is already reached, then the UI should display it clearly and disable all user actions.
> 2. It could also happen that the cap hasn't been reached when the user submits the deposit, but _then_ it will be reached. In that case, they need to be able to reclaim their Bitcoin.

For point 1 that part is complete it validates twice once when the user loads the page and a second time when the user clicks on next

For point 2, I think in that case the /deposit endpoint should assist in getting that info?

aldur:
> For point 2, I think in that case the /deposit endpoint should assist in getting that info?
Probably, but we will need to check with the sBTC team, can you ask them directly?

Also, I'd like to test this or see the UX. How are you testing it?

Jiloc:
The Emily API currently does not perform direct validation on deposit requests for the caps. It could potentially validate whether a single request exceeds the individual request cap, but it does not have sufficient information to enforce the total cap in edge cases. If a request violates either cap, it will be rejected by the signers. However, the signers don't update Emily when a request is rejected (it's possible, but for various operational reasons it wasn't implemented yet).

The most straightforward way for the bridge to check if a request was not accepted is to monitor the lock time expiration. If the lock time expires without the deposit being processed, users can assume their request was not included and reclaim their deposit

aldur:
> The Emily API currently does not perform direct validation on deposit requests for the caps. It could potentially validate whether a single request exceeds the individual request cap, but it does not have sufficient information to enforce the total cap in edge cases. If a request violates either cap, it will be rejected by the signers. However, the signers don't update Emily when a request is rejected (it's possible, but for various operational reasons it wasn't implemented yet).
> 
> The most straightforward way for the bridge to check if a request was not accepted is to monitor the lock time expiration. If the lock time expires without the deposit being processed, users can assume their request was not included and reclaim their deposit

This requires the bridge to keep local state though, which we are not currently doing.

aldur:
@hozzjss thanks for the video! I found it a bit confusing: the cap is on the _total_ amount of sBTC. So it is reasonable that if someone wants to send more than the total cap, we don't let them to that.

But also, you need to be considering the delta between the amount of sBTC that have been bridge and the cap. Do you do that?

The message should also be updated to reflect that.

package.json
```diff
@@ -25,8 +26,10 @@
     "process": "0.11.10",
     "react": "18.3.1",
     "react-dom": "18.3.1",
+    "server-only": "^0.0.1",
     "webpack": "5.95.0",
-    "yarn": "1.22.22"
+    "yarn": "1.22.22",
+    "yup": "^1.4.0"
```
aldur, line 32: Please keep dependencies pinned to the exact version.

package.json
```diff
@@ -15,6 +15,7 @@
     "@headlessui/react": "2.1.2",
     "@heroicons/react": "2.1.5",
     "@stacks/connect-react": "22.5.0",
+    "@tanstack/react-query": "^5.61.5",
```
aldur, line 18: Pin to exact version.

package.json
```diff
@@ -25,8 +26,10 @@
     "process": "0.11.10",
     "react": "18.3.1",
     "react-dom": "18.3.1",
+    "server-only": "^0.0.1",
```
aldur, line 29: Pin to exact version.

Jiloc:
> > The Emily API currently does not perform direct validation on deposit requests for the caps. It could potentially validate whether a single request exceeds the individual request cap, but it does not have sufficient information to enforce the total cap in edge cases. If a request violates either cap, it will be rejected by the signers. However, the signers don't update Emily when a request is rejected (it's possible, but for various operational reasons it wasn't implemented yet).
> > The most straightforward way for the bridge to check if a request was not accepted is to monitor the lock time expiration. If the lock time expires without the deposit being processed, users can assume their request was not included and reclaim their deposit
> 
> This requires the bridge to keep local state though, which we are not currently doing.

Mmm I see, that's annoying. The thing is, even if we add a check on Emily, we are only able to reject requests that exceed the per_deposit_cap. We cannot even check if the request would exceed the total_cap - current_supply because Emily doesn't have a stacks node to query for the current supply. While both checks can be done at the bridge level. 
If we had the signers update Emily about rejected requests, the bridge should still keep track of the request status via Emily - with the advantage that the rejection could maybe detected earlier than the lock time though.

aldur:
> > > The Emily API currently does not perform direct validation on deposit requests for the caps. It could potentially validate whether a single request exceeds the individual request cap, but it does not have sufficient information to enforce the total cap in edge cases. If a request violates either cap, it will be rejected by the signers. However, the signers don't update Emily when a request is rejected (it's possible, but for various operational reasons it wasn't implemented yet).
> > > The most straightforward way for the bridge to check if a request was not accepted is to monitor the lock time expiration. If the lock time expires without the deposit being processed, users can assume their request was not included and reclaim their deposit
> > 
> > 
> > This requires the bridge to keep local state though, which we are not currently doing.
> 
> Mmm I see, that's annoying. The thing is, even if we add a check on Emily, we are only able to reject requests that exceed the per_deposit_cap. We cannot even check if the request would exceed the total_cap - current_supply because Emily doesn't have a stacks node to query for the current supply. While both checks can be done at the bridge level. If we had the signers update Emily about rejected requests, the bridge should still keep track of the request status via Emily - with the advantage that the rejection could maybe detected earlier than the lock time though.

Let's discuss this later at standup, it seems like a design thing we need to fix.

hozzjss:
> @hozzjss thanks for the video! I found it a bit confusing: the cap is on the _total_ amount of sBTC. So it is reasonable that if someone wants to send more than the total cap, we don't let them to that.
> 
> But also, you need to be considering the delta between the amount of sBTC that have been bridge and the cap. Do you do that?
> 
> The message should also be updated to reflect that.
> 
> EDIT: looking at the code, I'd like to see what happens if `pegCap` is reached -- I don't think that is handled right now.

That's a good point `pegCap` should be checked too, sBTC - fee = amount, so basically that should be the amount compared to the peg, I just gotta find out where that fee is set, also shouldn't we show the user the sBTC they will receive here too?

aldur:
As discussed at standup, you can check the SC for the circulating supply (instead of asking Emily).

aldur (approved):

PR #36, opened 2024-11-29
feat - add lock time to env config set up

setbern:
Ensure the lock time is set up in the env variable


hozzjss (approved):

src/util/atoms.ts
```diff
@@ -15,6 +15,7 @@ export const bridgeConfigAtom = atom<BridgeConfig>({
   SBTC_CONTRACT_DEPLOYER: undefined,
   SIGNER_AGGREGATE_KEY: undefined,
   BANNER_CONTENT: undefined,
+  RECLAIM_LOCK_TIME: 144,
```
aldur, line 18: Why is the default duplicated here and in `env.ts`?

hozzjss: env.ts is only accessible to the server

setbern: Removed duplication, added a fallback in `DepositFlowConfirm` in case env var is ever undefined

PR #39, opened 2024-12-04
use sats internally and btc in UI

hozzjss:
Use the following template to create your pull request

## Description

Sats internally and BTC externally

https://github.com/user-attachments/assets/46e3919b-65b6-4b6f-854d-80f867fd6f59

### DIFF
--- a/src/comps/Deposit.tsx
+++ b/src/comps/Deposit.tsx
@@ -76,11 +76,15 @@ type DepositFlowAmountProps = DepositFlowStepProps & {
 };
 const DepositFlowAmount = ({ setStep, setAmount }: DepositFlowAmountProps) => {
   const { currentCap, isWithinDepositLimits, isLoading } = useMintCaps();
+  const minDepositAmount = 100_000 / 1e8;
   const validationSchema = yup.object({
     amount: yup
       .number()
       // dust amount is in sats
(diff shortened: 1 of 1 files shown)


src/comps/Deposit.tsx
```diff
@@ -80,14 +80,15 @@ const DepositFlowAmount = ({ setStep, setAmount }: DepositFlowAmountProps) => {
     amount: yup
       .number()
       // dust amount is in sats
-      .min(10_000, `Minimum deposit amount is 10,000 sats`)
-      .max(currentCap, `Current deposit cap is ${currentCap} sats`)
+      .min(10_000 / 1e8, `Minimum deposit amount is ${10_000 / 1e8} BTC`)
```
aldur, line 83: Shouldn't this limit come from Emily rather than being hardened as 10_000 sats?

hozzjss: does emily provide a lower limit?

aldur: If it doesn't and you are adding it for UX, it should be at least 100_000 sats

hozzjss: got it!

aldur (approved):

PR #40, opened 2024-12-04
Chore/sbtc balance

hozzjss:
## Description

As a user I want to see my current sBTC balance in the UI

For details refer to issue https://github.com/stacks-network/sbtc/issues/1022

## Type of Change
- Other 

No

## Are documentation updates required?
 No

## Testing information
Log in with your leather wallet and checkout your balance

### DIFF
--- a/src/comps/ConnectWallet.tsx
+++ b/src/comps/ConnectWallet.tsx
@@ -44,10 +44,8 @@ const ConnectWallet = ({ onClose }: ConnectWalletProps) => {
   const { WALLET_NETWORK } = useAtomValue(bridgeConfigAtom);
   const handleSelectWallet = async (wallet: WalletProvider) => {
     try {
-      let addresses = {
-        payment: "",
-        taproot: "",
-      };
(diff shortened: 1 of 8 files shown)


src/comps/ConnectWallet.tsx
```diff
@@ -44,10 +44,8 @@ const ConnectWallet = ({ onClose }: ConnectWalletProps) => {
   const { WALLET_NETWORK } = useAtomValue(bridgeConfigAtom);
   const handleSelectWallet = async (wallet: WalletProvider) => {
     try {
-      let addresses = {
-        payment: "",
-        taproot: "",
-      };
+      let addresses: Awaited<ReturnType<typeof getAddressesLeather>> | null =
+        null;
```
aldur, line 48: Will this work with Xverse?

aldur (approved):

src/comps/ConnectWallet.tsx
hozzjss: Yes this is an abstract interface I just didn't want to export the abstract type but it's worth fixing and using the abstract interface

PR #48, opened 2024-12-05
feat - add cors config to allow outside domains (specfically website)…

setbern:
Allow outside domains to hit the backend without having to go through a server.

This is for wallets and packages to be able to hit our backend directly 
- resolves [Issue](https://github.com/hirosystems/stacks.js/pull/1554)

Added a basic cors config and added the cors check to our API routes

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -26,6 +26,27 @@ const nextConfig = {
         ]
       : [];
   },
+  async headers() {
+    return [
+      {
+        // matching all API routes
+        source: "/api/:path*",
+        headers: [
+          { key: "Access-Control-Allow-Credentials", value: "true" },
+          { key: "Access-Control-Allow-Origin", value: "*" }, // replace this your actual origin
+          {
+            key: "Access-Control-Allow-Methods",
+            value: "GET,DELETE,PATCH,POST,PUT",
+          },
+          {
+            key: "Access-Control-Allow-Headers",
+            value:
+              "X-CSRF-Token, X-Requested-With, Accept, Accept-Version, Content-Length, Content-MD5, Content-Type, Date, X-Api-Version",
+          },
+        ],
+      },
+    ];
+  },
 };
 
 module.exports = nextConfig;


package.json
```diff
@@ -14,11 +14,18 @@
...
+    "@scure/bip32": "1.3.3",
     "@stacks/transactions": "7.0.2",
     "@tanstack/react-query": "5.61.5",
+    "@types/cors": "^2.8.17",
     "add": "2.0.6",
     "babel-loader": "9.2.1",
+    "bitcoinjs-lib": "7.0.0-rc.0",
+    "bs58check": "4.0.0",
     "c32check": "2.0.0",
+    "cors": "^2.8.5",
```
hozzjss, line 27: pin this package

package.json
```diff
@@ -14,11 +14,18 @@
     "@bitcoinerlab/secp256k1": "1.1.1",
     "@headlessui/react": "2.1.2",
     "@heroicons/react": "2.1.5",
+    "@noble/hashes": "1.5.0",
+    "@scure/bip32": "1.3.3",
     "@stacks/transactions": "7.0.2",
     "@tanstack/react-query": "5.61.5",
+    "@types/cors": "^2.8.17",
```
hozzjss, line 21: pin this package

src/util/cors.ts
```diff
@@ -0,0 +1,24 @@
...
+const corsConfig = Cors({
+  methods: ["GET", "POST", "PUT", "DELETE", "OPTIONS"],
+  origin: "*",
+});
+
+export function runMiddleware(
+  req: NextRequest,
+  res: NextResponse,
+  fn: Function,
+): Promise<void> {
```
hozzjss, line 13: maybe add a check here that specifies that this is only allowed in sBTC testnet otherwise returns the res as is, if I understand this correctly in mainnet there will be apis that replace the proxy stuff

hozzjss (approved):

PR #56, opened 2024-12-09
Bern reclaim

setbern:
Solves 
- https://github.com/stacks-network/sbtc/issues/936
- https://github.com/stacks-network/sbtc/issues/941

Allow the user to reclaim a deposit request that was not swept by the signers.

This is not a mainnet ready deposit as the reclaim script is very bare bone and need bitcoin script improvement.

Specifically what prevents this from mainnet launch is adding `OPCHECKSIG` which checks a pubkey against a user signature.

Right now for testing purposes the reclaim script can be redeemed by anyone since we only have a basic time lock check


src/comps/Header.tsx
setbern, line 1: This need to be removed and hooked up to the env var

next.config.js
```diff
@@ -1,5 +1,13 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  async rewrites() {
+    return [
+      {
+        source: "/api/proxy/:path*",
+        destination: "https://beta.sbtc-mempool.tech/api/proxy/:path*",
+      },
+    ];
+  },
```
hozzjss, line 10: remove this

src/app/api/emilyDeposit/route.ts
```diff
@@ -78,6 +78,8 @@ export async function GET(req: NextRequest) {
     // Construct the URL for the external request
     const url = `${env.EMILY_URL}/deposit/${bitcoinTxid}/${vout}`;
 
+    console.log("url", url);
+
```
hozzjss, line 82: remove this

src/app/api/tx/route.ts
```diff
@@ -0,0 +1,43 @@
...
+
+//https://mempool.bitcoin.regtest.hiro.so/api/tx/add6c6aa331a134dbb82227ebae5dd527e6f6713bb8a6c9c98ec01195c44d83d
+
+export async function GET(req: NextRequest) {
+  try {
+    const searchParams = req.nextUrl.searchParams;
+    const txId = searchParams.get("txId");
+
+    const testNetUrl = "https://mempool.bitcoin.regtest.hiro.so/api/tx/";
+    const mainnetUrl = "https://mempool.space/api/tx/";
```
hozzjss, line 12: hardcoded apis use env or actions

src/comps/Deposit.tsx
```diff
@@ -240,12 +240,24 @@ const DepositFlowConfirm = ({
...
+        return;
+      }
+
+      // const reclaimPublicKey = paymentAddress.publicKey;
+
       // Parse lockTime from env variable
-      const parsedLockTime = parseInt(lockTime || "144");
+      //const parsedLockTime = parseInt(lockTime || "144");
+      // FIXME: need to get the locktime from the env variable
+      const parsedLockTime = 2;
```
hozzjss, line 258: use env var here

src/comps/Header.tsx
```diff
@@ -63,6 +62,10 @@ const Header = ({ config }: { config: BridgeConfig }) => {
       </>
     );
   };
+
+  if (!config) {
+    return null;
+  }
```
hozzjss, line 68: this will never happen config will always be passed

src/comps/Header.tsx
```diff
@@ -22,9 +22,8 @@ import Metrics from "./ui/metrics";
 // converting to lower case to avoid case sensitive issue
 
 const Header = ({ config }: { config: BridgeConfig }) => {
-  const isTestnet =
-    config.WALLET_NETWORK?.toLowerCase() === "sbtcTestnet".toLowerCase();
-
+  //const isTestnet = config.WALLET_NETWORK?.toLowerCase() === "sbtcTestnet".toLowerCase();
+  const isTestnet = true;
```
hozzjss, line 26: This is for the faucet it will only work in the sbtcTestnet revert back to the way it was

src/comps/ReclaimManager.tsx
```diff
@@ -0,0 +1,534 @@
...
+
+  const currentStep = useMemo(() => {
+    const steps = [ReclaimStatus.Pending, ReclaimStatus.Completed];
+    return steps.findIndex((step) => step === status);
+  }, [status]);
+
+  const memepoolUrl = useMemo(() => {
+    const testNetUrl = "https://mempool.bitcoin.regtest.hiro.so/tx/";
+    const mainnetUrl = "https://mempool.space/tx/";
+
```
hozzjss, line 482: Must not be hardcoded read from env or use actions

src/util/index.ts
```diff
@@ -1,3 +1,9 @@
 export function classNames(...classes: any[]) {
   return classes.filter(Boolean).join(" ");
 }
+
+export function uint8ArrayToHexString(uint8Array: Uint8Array): string {
+  return Array.from(uint8Array)
+    .map((byte) => byte.toString(16).padStart(2, "0"))
+    .join("");
+}
```
hozzjss, line 9: use the @stacks/common bytesToHex instead

hozzjss (changes requested):

src/app/api/emilyDeposit/route.ts
setbern: fixed

src/comps/Header.tsx
setbern: Remvoed

src/comps/Header.tsx
setbern: Reverted back

src/util/index.ts
setbern: Updated to use @stacks/common

setbern:
This ensure user can properly reclaim their bitcoin after a failed deposit sweep (or a cap limit reach) after their the block locktime has been surpassed.

How is this done?

  const buildScript = script.compile([
    lockTimeEncoded,
    opcodes.OP_CHECKSEQUENCEVERIFY,
    opcodes.OP_DROP,
    schnorPublicKey,
    opcodes.OP_CHECKSIG,
  ]);

The above is the reclaim script used as the second leaf for the p2tr deposit transaction (first being the deposit script)

This script leverages 2 main op codes

    OP_CHECKSEQUENCEVERIFY to ensure that we can only spend this after the selected amount of blocks have been surpassed defined by lockTimeEncoded
    OP_CHECKSIG which is used to ensure that the signature provided (by the wallets) matches the schnorPublicKey associated with the transaction that made the deposit request initially,
    
   (this comment was originally written for PR (https://github.com/stacks-network/sbtc-bridge/pull/57)

.env.sample
```diff
@@ -1,6 +1,7 @@
 BITCOIND_URL=
 EMILY_URL=
 MEMPOOL_API_URL=https://leather.mempool.space/api
+MEMPOOL_URL=https://leather.mempool.space
```
aldur, line 4: We can't use these endpoints and they shouldn't be in the sample. Just refer to the standard mempool URLs (i.e. drop the `leather` prefix).

src/actions/bitcoinClient.ts
```diff
@@ -84,3 +84,22 @@ export const getRawTransaction = async (
   const result = await fetch(`${baseURL}/tx/${txid}`);
   return await result.json();
 };
+
+// Function to transmit a raw transaction to the network
+export const transmitRawTransaction = async (hex: string): Promise<any> => {
```
aldur, line 89: Is this still required now that we can get Leather to correctly send the tx?

src/comps/ReclaimManager.tsx
```diff
@@ -0,0 +1,582 @@
...
+  };
+  const fetchDepositInfoFromEmily = async () => {
+    try {
+      // get the depositTxId and outputIndex from the query params
+      const depositTxId = searchParams.get("depositTxId");
+      const outputIndex = searchParams.get("vout") || 0;
+
+      // we want to get the deposit info from Emily
+      /*
+        that means we need the txId of the deposit and optinoal the output index
```
aldur, line 224:
```suggestion
        that means we need the txId of the deposit and optionally the output index
```

src/env.ts
```diff
@@ -6,6 +6,8 @@ export const env = {
   EMILY_URL: process.env.EMILY_URL,
   MEMPOOL_API_URL:
     process.env.MEMPOOL_API_URL || "http://localhost:8083/api/v1/",
+  MEMPOOL_URL: process.env.MEMPOOL_URL,
```
aldur, line 9: Can't we just rely on a single env variable for MEMPOOL and derive the other from it?

src/util/depositRequest.ts
```diff
@@ -1,21 +1,16 @@
...
 import ecc from "@bitcoinerlab/secp256k1";
-const ECPair: ECPairAPI = ECPairFactory(ecc);
+// const ECPair: ECPairAPI = ECPairFactory(ecc);
 bitcoin.initEccLib(ecc);
 
-const NUMS_X_COORDINATE = new Uint8Array([
+export const NUMS_X_COORDINATE = new Uint8Array([
   0x50, 0x92, 0x9b, 0x74, 0xc1, 0xa0, 0x49, 0x54, 0xb7, 0x8b, 0x4b, 0x60, 0x35,
   0xe9, 0x7a, 0x5e, 0x07, 0x8a, 0x5a, 0x0f, 0x28, 0xec, 0x96, 0xd5, 0x47, 0xbf,
   0xee, 0x9a, 0xce, 0x80, 0x3a, 0xc0,
```
aldur, line 16: Where do these come from?

src/util/depositRequest.ts
```diff
@@ -1,21 +1,16 @@
...
-} from "@stacks/common";
+import { hexToBytes as hexToUint8Array } from "@stacks/common";
 
 import { Taptree } from "bitcoinjs-lib/src/types";
 
 import * as bip341 from "bitcoinjs-lib/src/payments/bip341";
 
 import ecc from "@bitcoinerlab/secp256k1";
-const ECPair: ECPairAPI = ECPairFactory(ecc);
+// const ECPair: ECPairAPI = ECPairFactory(ecc);
```
aldur, line 10: Please remove if unused.

src/util/depositRequest.ts
```diff
@@ -69,41 +64,29 @@ export const createDepositScript = (
...
-    lockTimeArray.length + opCheckSequenceVerify.length,
-  ); // Append additional script
-
   // Return the combined Uint8Array
   const buildScript = script.compile([
-    lockTimeArray,
+    lockTimeEncoded,
     opcodes.OP_CHECKSEQUENCEVERIFY,
+    opcodes.OP_DROP,
+    schnorPublicKey,
```
aldur, line 84:
```suggestion
    schnorrPublicKey,
```

src/util/depositRequest.ts
```diff
@@ -131,9 +115,8 @@ export const createDepositAddress = (
 
   //  Hash the leaf scripts using tapLeafHash
 
-  const reclaimScriptHash = bip341.tapleafHash({ output: reclaimScript });
-  const reclaimScriptHashHex = uint8ArrayToHexString(reclaimScriptHash);
-  console.log("reclaimScriptHashHex", reclaimScriptHashHex);
+  // const reclaimScriptHash = bip341.tapleafHash({ output: reclaimScript });
+  // const reclaimScriptHashHex = uint8ArrayToHexString(reclaimScriptHash);
```
aldur, line 119: Please cleanup comments.

src/util/reclaimHelper.tsx
```diff
@@ -0,0 +1,123 @@
...
+import * as bitcoin from "bitcoinjs-lib";
+
+import { Taptree } from "bitcoinjs-lib/src/types";
+
+import { hexToBytes as hexToUint8Array } from "@stacks/common";
+import { NUMS_X_COORDINATE } from "./depositRequest";
+
+export const finalizePsbt = (psbtHex: string) => {
+  try {
+    const network = bitcoin.networks.regtest;
```
aldur, line 11: How will this work on mainnet?

src/util/reclaimHelper.tsx
```diff
@@ -0,0 +1,123 @@
...
+    { output: uInt8DepositScript },
+    { output: uInt8ReclaimScript },
+  ]);
+
+  // Ensure Merkle is computed
+  if (!merkleeTree || !merkleeTree.hash) {
+    throw new Error("Failed to compute Merkle root.");
+  }
+
+  const network = bitcoin.networks.regtest;
```
aldur, line 75: How will this work for mainnet?

aldur (changes requested):

src/actions/bitcoinClient.ts
hozzjss: This is needed to broadcast the signed psbt here https://github.com/stacks-network/sbtc-bridge/blob/6e81b487a3451721b7418149015c7c968d86087d/src/comps/ReclaimManager.tsx#L406

src/hooks/use-reclaim-status.ts
```diff
@@ -0,0 +1,41 @@
...
+  // we'll need to fetch this from the bitcoin rpc to get the current status of the tx
+  const [reclaimStatus, setReclaimStatus] = useState<ReclaimStatus>(
+    ReclaimStatus.Pending,
+  );
+
+  useEffect(() => {
+    if (txId && reclaimStatus !== ReclaimStatus.Completed) {
+      // fetch the status of the reclaim tx from the bitcoin rpc
+      // and update the reclaimStatus
+      // setReclaimStatus(ReclaimStatus.Completed);
```
matteojug, line 20: nit: leftover

src/hooks/use-reclaim-status.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        let status = ReclaimStatus.Pending;
+
+        if (reclaimTx.status.confirmed) {
+          status = ReclaimStatus.Completed;
+          clearInterval(interval);
+        }
+
+        setReclaimStatus(status);
+      }, 1000);
```
matteojug, line 32: Can we move the polling interval (both for this and the deposit status) to an env? To be able to increase it easily if 1s end up being too frequent. Not necessarily in this PR.

src/util/depositRequest.ts
setbern: Was supplied by @djordon

src/app/api/proxy/[...proxy]/route.ts
```diff
@@ -80,7 +80,7 @@ export async function GET(req: NextRequest) {
     }
 
     // Proxy all other routes to the base proxy URL
-    const proxyUrl = `${MEMPOOL_API_URL}${path}`;
+    const proxyUrl = `${MEMPOOL_API_URL}/${path}`;
```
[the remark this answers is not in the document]
Comment (hozzjss) [reply src/app/api/proxy/[...proxy]/route.ts:73]: good idea

src/hooks/use-reclaim-status.ts
hozzjss: Good idea I'll create an issue

hozzjss: created https://github.com/stacks-network/sbtc/issues/1113

src/util/depositRequest.ts
aldur: Thanks, Daniel educated me: https://github.com/stacks-network/sbtc/blob/a3a927f759871440962d8f8066108e5b0af696a0/sbtc/src/lib.rs#L16-L33

djordon: Yeah these should be the same ones in

https://github.com/stacks-network/sbtc/blob/a3a927f759871440962d8f8066108e5b0af696a0/sbtc/src/lib.rs#L16-L33

Those bytes are the same ones in BIP-0341, and they are computed by taking the SHA256 of the generator point of the secp256k1 elliptic curve. I think the idea is that this point is verifiably random so it is unlikely that anyone knows the private key.

src/hooks/use-deposit-status.ts
```diff
@@ -30,5 +38,24 @@ export function useDepositStatus(txId: string) {
...
+      const info = await getRawTransaction(txId);
+      if (info.status.confirmed) {
+        const currentBlockHeight = await getCurrentBlockHeight();
+        const unlockBlock =
+          Number(RECLAIM_LOCK_TIME || 144) + info.status.block_height;
+        const isPastLockTime = currentBlockHeight > unlockBlock;
+        if (isPastLockTime) {
+          setTransferTxStatus(DepositStatus.Failed);
+          clearInterval(interval);
+        }
```
matteojug, line 52: We tested this branch on devnet, it seems that we do get in here once the deposit is unlocked but the page/component/thingy doesn't refresh and stays as:
image
Refreshing the page we do get the updated view

src/comps/ReclaimManager.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+      const btcAddress = getWalletAddress();
+
+      if (!btcAddress) {
+        return setShowWallet(true);
+      }
+
+      // FIXME: move to util or its own file
+      const unsignedTxHex = constructPsbtForReclaim({
+        depositAmount: amount * 1e8,
```
matteojug, line 353: We need to round this, otherwise when we try to convert it to bigint in`constructPsbtForReclaim` we may get an error like `The number 101200.00000000001 cannot be converted to a BigInt because it is not an integer`

src/env.ts
```diff
@@ -4,8 +4,11 @@ import { DefaultNetworkConfigurations } from "@leather.io/models";
 export const env = {
   BITCOIND_URL: process.env.BITCOIND_URL || "http://localhost:18443",
   EMILY_URL: process.env.EMILY_URL,
-  MEMPOOL_API_URL:
-    process.env.MEMPOOL_API_URL || "http://localhost:8083/api/v1/",
+  MEMPOOL_API_URL: (
+    process.env.MEMPOOL_API_URL || "http://localhost:8083/api"
+  ).replace(/\/$/, ""),
+  MEMPOOL_URL: process.env.MEMPOOL_URL,
```
matteojug, line 10: Can we do the same trailing slash replace also for `MEMPOOL_URL`?

src/comps/ReclaimManager.tsx
hozzjss: fixed

src/env.ts
hozzjss: smart, fixed!

matteojug (approved):
Tested in devenv

hozzjss (approved):
rubber stamp since I'm a contributor and a reviewer

aldur (approved):

PR #62, opened 2024-12-16
Fix/critical prod

hozzjss:
This fixes 3 critical bugs:
1. txid returned by proxy is no longer a stringified string.
2. emily url is not user editable any longer only set by the env
3. emily api key is no longer used or passed to emily for open apis

(edit by aldur) plus it makes it possible to set the minimum through an env variable


aldur:
Fixes:

- https://github.com/stacks-network/sbtc/issues/1139
- https://github.com/stacks-network/sbtc/issues/1138

aldur (commented):
Has this been tested again main? I'd wait to do that _before_ merging.

src/comps/Deposit.tsx
```diff
@@ -306,9 +307,10 @@ const DepositFlowConfirm = ({
         bitcoinTxOutputIndex: 0,
         reclaimScript: reclaimScriptHex,
         depositScript: depositScriptHexPreHash,
-        url: emilyUrl,
       };
 
+      console.log({ emilyReqPayloadClient: emilyReqPayload });
+
```
matteojug, line 313: nit: can we also log this (with empty `bitcoinTxid`) before calling leather/xverse? For the current issue we are having we would not get any log

hozzjss: done

hozzjss:
Fixes:
- https://github.com/stacks-network/sbtc/issues/1137
- https://github.com/stacks-network/sbtc/issues/1140
- https://github.com/stacks-network/sbtc/issues/1141

aldur:
We have tested `754e1c5bf4b77762e462bc93cb1f713a2624acd3` on mainnet bridge. I would merge this PR up to that point and ask you to open a new PR for the remaining commits.

PR #63, opened 2024-12-16
Add extra logs when sending a deposit and don't allow the user to exceed their balance

hozzjss:
Fixes:
- https://github.com/stacks-network/sbtc/issues/1137
- https://github.com/stacks-network/sbtc/issues/1141

### DIFF
--- a/src/comps/Deposit.tsx
+++ b/src/comps/Deposit.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 
 import { InformationCircleIcon } from "@heroicons/react/20/solid";
 
@@ -40,6 +40,8 @@ import { sendBTCLeather, sendBTCXverse } from "@/util/wallet-utils";
 import useMintCaps from "@/hooks/use-mint-caps";
 import { getAggregateKey } from "@/util/get-aggregate-key";
 import getBitcoinNetwork from "@/util/get-bitcoin-network";
+import { useQuery } from "@tanstack/react-query";
+import getBtcBalance from "@/actions/get-btc-balance";
 /*
   deposit flow has 3 steps
   1) enter amount you want to deposit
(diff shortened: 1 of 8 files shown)


src/comps/Deposit.tsx
```diff
@@ -75,25 +78,38 @@ type DepositFlowStepProps = {
...
-      .number()
-      // dust amount is in sats
-      .min(
-        minDepositAmount,
-        `Minimum deposit amount is ${minDepositAmount} BTC`,
-      )
-      .max(maxDepositAmount, `Current deposit cap is ${maxDepositAmount} BTC`)
-      .required(),
-  });
+  let minDepositAmount = perDepositMinimum / 1e8;
```
matteojug, line 91: nit: can be const

src/comps/Deposit.tsx
```diff
@@ -511,11 +547,27 @@ const DepositFlow = () => {
...
+    },
+    enabled: !!btcAddress,
+  });
   const renderStep = () => {
     switch (step) {
       case DEPOSIT_STEP.AMOUNT:
         return (
-          <DepositFlowAmount setAmount={setAmount} setStep={handleUpdateStep} />
+          <DepositFlowAmount
+            btcBalance={btcBalance || 0}
```
matteojug, line 567: IIUC defaulting to `0` means that if we are having troubles reaching out to the mempool api, users will not be able to use the bridge. If so, should we use a `MAX_MONEY` default instead (or disable validation if we cannot get the balance)? Alternatively, we could change the check wrt the user balance to be just a fyi kind of thing, not blocking the flow -- to that if we/api mess up the balance computation, the user can still go ahead with the process.

src/hooks/use-mint-caps.ts
```diff
@@ -38,9 +38,11 @@ export default function useMintCaps() {
       if (!currentEmilyLimitsData || !currentSbtcSupplyData) {
         return false;
       }
-      const withinMintCap =
-        currentEmilyLimitsData.pegCap >
+      const supplyPlusDeposit =
         Number(currentSbtcSupplyData.value.value) + amount;
+      const withinMintCap =
+        amount > currentEmilyLimitsData.perDepositMinimum &&
+        supplyPlusDeposit < currentEmilyLimitsData.pegCap;
```
matteojug, line 45: Should we use `>=` and `<=` here?

src/comps/Deposit.tsx
```diff
@@ -281,6 +297,15 @@ const DepositFlowConfirm = ({
...
           network: walletNetwork,
         };
+        console.log({
+          preSendParams: {
+            bitcoinTxid: txId,
+            bitcoinTxOutputIndex: 0,
+            reclaimScript: reclaimScriptHex,
+            depositScript: depositScriptHexPreHash,
+          },
+        });
```
aldur, line 307: Can you make this a JSON? Thanks!

matteojug (approved):

PR #65, opened 2024-12-16
feat: Live chat and live support link

setbern:
Add Live chat for customer support


src/app/layout.tsx
```diff
@@ -27,6 +27,50 @@ export default async function RootLayout({
...
+                  off:function(){i(["off",[].slice.call(arguments)])},
+                  get:function(){
+                    if(!e._h)throw new Error("[LiveChatWidget] You can't use getters before load.");
+                    return i(["get",[].slice.call(arguments)])
+                  },
+                  call:function(){i(["call",[].slice.call(arguments)])},
+                  init:function(){
+                    var n=t.createElement("script");
+                    n.async=!0,n.type="text/javascript",
+                    n.src="https://cdn.livechatinc.com/tracking.js",
```
aldur, line 51: @matteojug using your JS-fu: this bit was blocked by my ad-blocker, so I did not see the support icon. That's unfortunate. But it seems like this is required for the chat thing to work, it's not just for "tracking" as the name suggests. Is this correct?

PR #71, opened 2024-12-17
replace by fee handling for deposits

hozzjss:
This checks if the deposit tx exists and if it doesn't it fetches if it has been rbf'ed and if so redirect to the rbf'ed tx after notifying emily

### DIFF
--- a/src/actions/bitcoinClient.ts
+++ b/src/actions/bitcoinClient.ts
@@ -80,8 +80,16 @@ export const scanTxOutSet = async (
 // Function to get a raw transaction
 export const getRawTransaction = async (
   txid: string,
-): Promise<BitcoinTransactionResponse> => {
+): Promise<BitcoinTransactionResponse | null> => {
   const result = await fetch(`${env.MEMPOOL_API_URL}/tx/${txid}`);
+  if (result.status === 404) {
+    return null;
+  }
+  return await result.json();
+};
+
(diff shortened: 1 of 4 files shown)


src/actions/bitcoinClient.ts
```diff
@@ -80,8 +80,16 @@ export const scanTxOutSet = async (
...
+): Promise<BitcoinTransactionResponse | null> => {
   const result = await fetch(`${env.MEMPOOL_API_URL}/tx/${txid}`);
+  if (result.status === 404) {
+    return null;
+  }
+  return await result.json();
+};
+
+export const getTxRbf = async (txid: string): Promise<boolean> => {
+  const result = await fetch(`${env.MEMPOOL_API_URL}/v1/tx/${txid}/rbf`);
```
matteojug, line 92: We don't have the `v1` part above, is it expected?

matteojug, line 92: Should this also have the `result.status` guard?

matteojug (approved):

src/actions/bitcoinClient.ts
hozzjss: yes for this one it is

hozzjss: actually no, and I'm being optimistic here tbh, basically the user can do rbf which could cancel the transaction which breaks everything, that case is not handled

PR #73, opened 2024-12-17
Hotfix emily lost tx

setbern:
Ensure we can rebuild deposit and reclaim script to inform emily in the scenario where emily did not get the info the first time

in order for this to work we need two inputs as seach param

recover?txId=740e919c2718dded310cce09f8c31ddf4c0c84be808dec5ba605dd037041b8de&stxAddress=ST1S2ZTV7QVAYBRJVB85FHXE7P8PZZHXVCC40BA8R

`txId` the btc transaction id that was created
`stxAddress` the address where the sBTC is intended to be sent to


src/comps/recover.tsx
```diff
@@ -0,0 +1,233 @@
...
+        message: `Request successful`,
+      });
+      setShowStepper(true);
+    } catch (err: any) {
+      throw new Error(err);
+    }
+  };
+
+  const renderTxId = () => {
+    const txId = searchParams.get("txId") || "";
```
matteojug, line 173: Given the funny behaviour in chrome, should we switch to lowercase param?

src/comps/recover.tsx
```diff
@@ -0,0 +1,233 @@
...
+  if (showStepper) {
+    return <RecoverReview txId={searchParams.get("txId") || ""} />;
+  }
+
+  return (
+    <>
+      <FlowContainer>
+        <>
+          <div className="w-full flex flex-row items-center justify-between">
+            <Heading>Recover Lost Transaction</Heading>
```
matteojug, line 188: Should we add some more details about what this is about?

src/comps/recover.tsx
```diff
@@ -0,0 +1,233 @@
...
+      const reclaimScriptHex = uint8ArrayToHexString(reclaimScript);
+
+      const signerUint8Array = hexToUint8Array(signersAggregatePubKey!);
+
+      const depositScript = Buffer.from(
+        createDepositScript(signerUint8Array, maxFee, serializedAddress),
+      );
+      // convert buffer to hex
+      const depositScriptHexPreHash = uint8ArrayToHexString(depositScript);
+
```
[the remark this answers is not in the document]
aldur: +1 wanted to do the same on my script

src/comps/recover-review.tsx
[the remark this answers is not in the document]
setbern: You're right, was removed and now links back to that stepper view

src/comps/recover.tsx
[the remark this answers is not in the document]
setbern: Added a check that recreates the p2tr address and fetches the vout from the bitcoin transaction passed.

Only passes if the address recreated from recover script match a vout in the fetch bitcoin transaction

src/comps/recover.tsx
setbern: Added a check to ensure it get from both cases

src/comps/recover.tsx
setbern: Added `In the rare situation where your transaction was lost, use this tool to ensure It's placed back on track within our system` - don't love it but a placeholder till coms team looks it over

hozzjss (approved):
LGTM, however I would use inputs instead of URL for UX but as a bandaid all looks good

src/comps/recover.tsx
```diff
@@ -0,0 +1,333 @@
...
+      });
+
+      const params = new URLSearchParams();
+
+      params.set("txId", txId);
+      params.set("step", String(1));
+      params.set("amount", String(amount));
+      params.set("stxaddress", String(stxAddress));
+
+      router.push(`${pathname}?${params.toString()}`);
```
matteojug, line 273: Is this redirecting to the correct page?

src/comps/recover.tsx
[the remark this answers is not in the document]
matteojug: You added the check in `getTransactionAmount`, but when the user clicks on recover it triggers this fn (`generateScripts`), so the check is skipped. Am I missing something?

hozzjss:
@aldur @matteojug Is this still needed?

aldur:
I think this is part of the reclaim page now?

PR #75, opened 2024-12-18
add infinite retry to notify emily

hozzjss:


src/util/use-emily-deposit.ts
```diff
@@ -0,0 +1,38 @@
...
+    }) => {
+      const res = await fetch("/api/emilyDeposit", {
+        method: "POST",
+        headers: {
+          "Content-Type": "application/json",
+        },
+        body: JSON.stringify(params),
+      });
+      if (!res.ok) {
+        if (failureCount > 2) {
```
matteojug, line 24: qq: the `>2` is just to avoid spamming users too soon?

src/util/use-emily-deposit.ts
```diff
@@ -0,0 +1,38 @@
+import { NotificationStatusType } from "@/comps/Notifications";
+import { useNotifications } from "@/hooks/use-notifications";
+import { useMutation } from "@tanstack/react-query";
+
+const expBackoff = (attempt: number) =>
+  Math.min(attempt > 1 ? 2 ** attempt * 1000 : 1000, 30 * 1000);
```
matteojug, line 6: nit: I guess this comes from the doc, but why do we need the ternary? also, nit
```suggestion
  Math.min(2 ** attempt, 30) * 1000;
```

src/comps/Deposit.tsx
```diff
@@ -412,7 +408,12 @@ const DepositFlowConfirm = ({
           <SecondaryButton onClick={() => handlePrevClick()}>
             PREV
           </SecondaryButton>
-          <PrimaryButton onClick={() => handleNextClick()}>NEXT</PrimaryButton>
+          <PrimaryButton
+            disabled={isPendingNotifyEmily}
```
matteojug, line 412: Should we also disable `PREV` if there's a deposit in flight we are tracking?

hozzjss: makes sense

src/util/use-emily-deposit.ts
hozzjss: yes

src/comps/Deposit.tsx
hozzjss: done

aldur (approved):

PR #76, opened 2024-12-18
Asigna integration

hozzjss:
## Description

Asigna BTC Connect integration to sBTC bridge frontend to enable BTC deposits from Asigna directly to Stacks as sBTC.

## Type of Change
- New feature

Not a breaking change

By: @jeffrystone


src/util/wallet-utils/src/checkWallets.ts
```diff
@@ -21,5 +22,6 @@ export const checkAvailableWallets: () => Promise<{
   return {
     leather: isLeather,
     xverse: isXverse,
+    asigna: window.top !== window.self,
```
hozzjss, line 25: this check is waaay too broad can we have a checker in the asigna connect library that might include like a checker for the host maybe 

image

That's the only blocker for now

src/util/wallet-utils/src/sign-psbt.ts
```diff
@@ -34,7 +32,9 @@ export async function signPSBTXverse({ hex, address }: SignPSBTParams) {
...
   }
 
-  return bytesToHex(
-    Uint8Array.from(atob(response.result.psbt), (c) => c.charCodeAt(0)),
-  );
+  return base64ToHex(response.result.psbt);
+}
+
+export function base64ToHex(base64: string) {
+  return bytesToHex(Uint8Array.from(atob(base64), (c) => c.charCodeAt(0)));
```
matteojug, line 39: nit: there's also `Uint8Array.fromBase64` (which they say it's preferred over `atob`)

matteojug (approved):

src/util/reclaimHelper.tsx
```diff
@@ -10,13 +10,10 @@ import { NUMS_X_COORDINATE } from "./depositRequest";
...
 
 export const finalizePsbt = (
-  psbtHex: string,
+  psbt: bitcoin.Psbt,
   walletNetwork?: DefaultNetworkConfigurations,
 ) => {
   try {
-    const network = getBitcoinNetwork(walletNetwork);
-
-    const psbt = bitcoin.Psbt.fromHex(psbtHex, { network });
```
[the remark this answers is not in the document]
hozzjss: good point removed!

src/util/wallet-utils/src/sign-psbt.ts
hozzjss: I would have loved to use this as a native API however its support is very very limited unfortunately 
image

PR #83, opened 2024-12-20
add env var for polling ops

hozzjss:
@matteojug earlier asked for this, this should add an interval in the env for interval operations, this would apply to the deposit status and the reclaim statuses


matteojug (approved):
🙏

PR #87, opened 2025-01-21
feat: switch to @stacks/connect

janniks:
- Remove the custom ConnectWallet component and switch to `@stacks/connect`
- Simplifies a lot of things in favor of two hooks: `useConnectWallet` and `useDisconnectWallet`


hozzjss:
Looks promising tested connecting the wallet and the provider key under window is preserved, but would wait till this comes out of alpha

hozzjss:
@janniks we decided it's not the time yet since the compatibility isn't 100% there yet we can try this again later

PR #89, opened 2025-01-23
switch to derived NUMS

MCJOHN974:
## Description

Currently we use NUMS internal key for taproot reclaim transactions. Unfortunately, Ledger don't allow users to sign trunsactions with not-deriven internal key, and Ledger users can't do reclaim. This PR switches from NUMS to key, derived from NUMS with path "0/0"

## Checklist
- [x] Code is commented where needed
- [x] Unit test coverage for new or modified code paths
- [x] `npm run test` passes
- [ ] Changelog is updated
- [x] Performed self-review on my code


src/util/depositRequest.ts
```diff
@@ -16,6 +16,13 @@ export const NUMS_X_COORDINATE = new Uint8Array([
   0xee, 0x9a, 0xce, 0x80, 0x3a, 0xc0,
 ]);
```
aldur, line 17: Is this still used anywhere?

src/util/depositRequest.ts
```diff
@@ -16,6 +16,13 @@ export const NUMS_X_COORDINATE = new Uint8Array([
...
 
+export const DERIVED_NUMS_X_COORDINATE = new Uint8Array([
+  0x4a ,0x30 ,0xb2 ,0xe4 ,0x61 ,0xb2 ,0x80, 0xc0,
+  0xb1, 0x3a, 0x03, 0x79, 0x90, 0x96, 0xab, 0x12,
+  0x56, 0x58, 0x91, 0x53, 0xfc, 0x4b, 0x9c, 0x8c,
+  0xea, 0xd1, 0x6d, 0xc0, 0xb6, 0x42, 0x06, 0x97,
+]);
```
aldur, line 24: Please fix formatting

aldur:
Note for ourselves: this will need to be merged after https://github.com/stacks-network/sbtc/pull/1255

src/util/depositRequest.ts
aldur: I think it's OK to delete it

aldur: Yeah I did comment it's OK to delete it?

aldur:
Closing this (since we can't switch to a derived pub-key for the signer aggregate key).

PR #90, opened 2025-01-30
add transactionHex to parameters when notifying emily

hozzjss:
make sure to fail the whole request and retry if tx not found yet

### DIFF
--- a/src/app/api/emilyDeposit/route.ts
+++ b/src/app/api/emilyDeposit/route.ts
@@ -12,11 +12,11 @@ interface CreateDepositRequestBody {
 export async function POST(req: NextRequest) {
   try {
     const body: CreateDepositRequestBody = await req.json();
-    const transactionHex = await getTransactionHex(body.bitcoinTxid);
-
     console.log({
       emilyReqPayloadInitialServer: JSON.stringify(body),
     });
+    const transactionHex = await getTransactionHex(body.bitcoinTxid);
+
     const paramsBody = {
       bitcoinTxid: body.bitcoinTxid,
       bitcoinTxOutputIndex: body.bitcoinTxOutputIndex,
--- a/src/app/api/emilyDeposit/route.ts
+++ b/src/app/api/emilyDeposit/route.ts
@@ -6,6 +6,7 @@ interface CreateDepositRequestBody {
   bitcoinTxOutputIndex: number;
   reclaimScript: string;
   depositScript: string;
+  transactionHex: string;
 }
 
 export async function POST(req: NextRequest) {
@@ -17,6 +18,7 @@ export async function POST(req: NextRequest) {
       bitcoinTxOutputIndex: body.bitcoinTxOutputIndex,
       reclaimScript: body.reclaimScript,
       depositScript: body.depositScript,
+      transactionHex: body.transactionHex,
     };
     console.log({ emilyReqPayloadServer: JSON.stringify(paramsBody) });
 
--- a/src/app/api/emilyDeposit/route.ts
+++ b/src/app/api/emilyDeposit/route.ts
@@ -1,3 +1,4 @@
+import { getTransactionHex } from "@/actions/bitcoinClient";
 import { env } from "@/env";
 import { NextRequest, NextResponse } from "next/server";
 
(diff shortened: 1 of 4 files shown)


hozzjss:
added @aldur @matteojug couldn't add @Jiloc not sure why

aldur:
Added @Jiloc (had to first invite him to the repo)

src/comps/Deposit.tsx
```diff
@@ -352,8 +352,6 @@ const DepositFlowConfirm = ({
         depositScript: depositScriptHexPreHash,
       };
 
-      console.log({ emilyReqPayloadClient: JSON.stringify(emilyReqPayload) });
```
aldur, line 355: I actually would like to keep this -- it allows to easily see what was supposed to be sent to Emily in case of errors.

Jiloc: I would still keep a log here, mainly because the new log happens after the `getTransactionHex`. If we'll have any problems with mempool, we may never log the deposit info. 

Maybe just rename this into something like  `emilyReqPayloadInitial` ?

src/util/use-emily-deposit.ts
```diff
@@ -13,12 +13,18 @@ export const useEmilyDeposit = () => {
       reclaimScript: string;
       depositScript: string;
     }) => {
+      const transactionHex = await getTransactionHex(params.bitcoinTxid);
+      const payload = {
+        ...params,
+        transactionHex,
+      };
+      console.log({ emilyReqPayloadClient: JSON.stringify(payload) });
```
Jiloc, line 21: should we move this into a separate mutation so that you won't repeat the logging and `getTransactionHex` calls when Emily fails?

src/comps/Deposit.tsx
hozzjss: okay sure

src/util/use-emily-deposit.ts
hozzjss: I moved it to the server so that failure is impossible unless the dedicated mempool api we're connected to fails in this instance we'll keep retrying since it fails

Jiloc: Maybe I am missing some background, when you say that it cannot fail, why is that? Can't our mempool api go down?

src/app/api/emilyDeposit/route.ts
```diff
@@ -11,12 +12,17 @@ interface CreateDepositRequestBody {
 export async function POST(req: NextRequest) {
   try {
     const body: CreateDepositRequestBody = await req.json();
+    const transactionHex = await getTransactionHex(body.bitcoinTxid);
 
+    console.log({
+      emilyReqPayloadInitialServer: JSON.stringify(body),
+    });
```
Jiloc, line 19: I believe this should go before the `const transactionHex = await getTransactionHex(body.bitcoinTxid);` line right?

src/util/use-emily-deposit.ts
Jiloc: I had a chat with @matteojug and he made me realize that we are using an external service in production:
```
MEMPOOL_API_URL: https://bitcoinltwolabs.mempool.space/api
PUBLIC_MEMPOOL_URL: https://bitcoinltwolabs.mempool.space/
``` 

So the call may fail. I'd still place it in a new mutation or compute it and cache it directly in the one where you call Emily.

src/app/api/emilyDeposit/route.ts
```diff
@@ -11,12 +12,17 @@ interface CreateDepositRequestBody {
 export async function POST(req: NextRequest) {
   try {
     const body: CreateDepositRequestBody = await req.json();
+    const transactionHex = await getTransactionHex(body.bitcoinTxid);
```
matteojug, line 15: If `getTransactionHex` fails, will this return a "nice" error to the api caller or just a generic 500?

src/app/api/emilyDeposit/route.ts
hozzjss: get transaction hex is now done on the server so this should log everything but the raw transaction hex

src/app/api/emilyDeposit/route.ts
hozzjss: it should return the error from the response

src/util/use-emily-deposit.ts
hozzjss: if the call to mempool fails the whole request fails and has to be retried from the frontend since the tx hex is a required parameter

PR #92, opened 2025-02-06
Chore -  Split reskin (v2) into seperate sub route so v1 can still stand while we push updates to v2

setbern:
Goal of this pr is to have V1 bridge live with V2. 

This is accomplished by having v2 live in a sub route `/reclaim`, v1 lives as normal in `/`.
There should be no UI files being references inside /reclaim to v1 components.

On the other hand as much business logic should be reused in order to prevent duplication of work and possible mistakes. 

This pr aims to have both version of the bridge coexist, does not promise all bugs solved on v2 but it should not effect in the slightest v1 bridge. This is **extremely** important to enforce, If any change to v1 bridge is found please flag asap. From my testing and PR review there seem to be only one major v1 change.
- `LayoutClient` has custom logic to render v1 or v2 header and footers.


setbern:
Closing this for the introduction of the new restructure 
https://github.com/stacks-network/sbtc/issues/1370

PR #94, opened 2025-02-20
Chore/social banner

hozzjss:
Adds a new BRIDGE_APP_URL env var that is basically the url of the bridge, needed for external social websites to load the banner


aldur (approved):
Looks good, just let @fabergat know that you have added a new env variable.

PR #99, opened 2025-02-24
Feat/add fordefi

hozzjss:
@janniks 's fordefi integration with the old method copied over to a local branch


src/util/wallet-utils/src/getAddress.ts
```diff
@@ -100,6 +103,27 @@ export const getAddressesXverse: getAddresses = async (params) => {
...
+  const response = await request(
+    "wallet_connect",
+    {
+      addresses: [
+        AddressPurpose.Ordinals,
+        AddressPurpose.Payment,
+        AddressPurpose.Stacks,
+      ],
+    },
+    FORDEFI_PROVIDER_ID,
```
matteojug, line 116: Do we need to specify xverse provider id as well in `getAddressesXverse` above?

src/util/wallet-utils/src/getAddress.ts
```diff
@@ -59,9 +62,9 @@ export function getWalletAddresses(
   }
 
   const stacks = getAddressByPurpose(response, AddressPurpose.Stacks);
-  if (!stacks) {
-    throw new Error("Stacks address not found");
-  }
+  // if (!stacks) {
+  //   throw new Error("Stacks address not found");
+  // }
```
matteojug, line 67: It also seems we are not using `taproot` around, should we comment out that check as well?

src/util/wallet-utils/src/getAddress.ts
```diff
@@ -72,8 +75,8 @@ export function getWalletAddresses(
       publicKey: payment.publicKey,
     },
     stacks: {
-      address: stacks.address,
-      publicKey: stacks.publicKey,
+      address: stacks?.address || "",
+      publicKey: stacks?.publicKey || "",
```
matteojug, line 79: If this is not set, are we just going to get `...` in `SBTCBalance` or something bad will happen?

src/util/wallet-utils/src/sign-psbt.ts
```diff
@@ -36,6 +39,21 @@ export async function signPSBTXverse({ hex, address }: SignPSBTParams) {
   return base64ToHex(response.result.psbt);
 }
 
+export async function signPSBTFordefi({ hex, address }: SignPSBTParams) {
+  const provider = await getFordefiBTCProviderOrThrow();
+  const response = await provider.unisatProvider.signOrSignAndSendPsbt(hex, {
```
matteojug, line 44: sanity check: is that `unisatProvider` correct here? Also I've got no clue where to look for `signOrSignAndSendPsbt`

src/util/wallet-utils/src/getAddress.ts
hozzjss: No it's the default

src/util/wallet-utils/src/getAddress.ts
hozzjss: good point!

src/util/wallet-utils/src/sign-psbt.ts
hozzjss: neither do I, no docs for this behavior unfortunately, I'm in the dark here but this seems to be working

src/util/wallet-utils/src/getAddress.ts
hozzjss: fixed

matteojug (approved):

PR #101, opened 2025-02-27
All stacks api calls to be through server actions and add stacks api env

hozzjss:
All stacks api calls now go through the server provided STACKS_API_URL
this env var would be either:

```sh
# Devenv
STACKS_API_URL=http://localhost:3999
# Devnet
STACKS_API_URL=https://api.testnet.hiro.so
# Mainnet
STACKS_API_URL=https://api.hiro.so
# custom deployment e.g. docker
STACKS_API_URL=not sure how docker does it lol
```


djordon (commented):
You missed one in this file https://github.com/stacks-network/sbtc-bridge/blob/chore/stx-api-env/src/actions/get-current-sbtc-supply.ts

hozzjss:
> You missed one in this file https://github.com/stacks-network/sbtc-bridge/blob/chore/stx-api-env/src/actions/get-current-sbtc-supply.ts

Fixed

djordon (approved):
This fixes the issue for me. Still no smooth sailing in our local devenv, but I'll open another ticket for any other issues.

PR #112, opened 2025-03-05
Chore/support all bitcoin addresses for withdrawals

hozzjss:
This would support all addresses for withdrawals decoding any bitcoin address

### DIFF
--- a/src/comps/temp-withdraw.tsx
+++ b/src/comps/temp-withdraw.tsx
@@ -8,18 +8,13 @@ import LandingAnimation from "./core/LandingAnimation";
 import { DefaultNetworkConfigurations } from "@leather.io/models";
 import { useRouter } from "next/navigation";
 
-import { bytesToHex, hexToBytes } from "@stacks/common";
-
-import * as bitcoin from "bitcoinjs-lib";
 import { useAtomValue } from "jotai";
 import { bridgeConfigAtom, walletInfoAtom, WalletProvider } from "@/util/atoms";
 import {
-  bufferCV,
+  Cl,
   makeUnsignedContractCall,
-  PostConditionMode,
+  Pc,
   serializeTransaction,
-  tupleCV,
-  uintCV,
 } from "@stacks/transactions";
 import { serverBroadcastTx } from "@/actions/server-broadcast-tx";
 import {
@@ -29,6 +24,8 @@ import {
   callContractXverse,
 } from "@/util/wallet-utils";
 import { getStacksNetwork } from "@/util/get-stacks-network";
+import { validate, Network } from "bitcoin-address-validation";
+import { decodeBitcoinAddress } from "@/util/decode-bitcoin-address";
 
 const decodeBitcoinAddressToClarityRecipient = (
   address: string,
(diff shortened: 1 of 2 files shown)


src/comps/temp-withdraw.tsx
```diff
@@ -29,65 +24,25 @@ import {
...
-      throw new Error(
-        `Unrecognized witnessVersion=${witnessVersion}, length=${witnessProgram.length}`,
-      );
-    }
+  const addressInfo = getAddressInfo(address);
+  let addressNetwork = isMainnet ? Network.mainnet : Network.regtest;
+  // bech32 is different between all three networks
+  // base58 is same between regtest and testnet
+  const isRegtestBase58 = !addressInfo.bech32 && !isMainnet;
+  addressNetwork = isRegtestBase58 ? Network.testnet : addressNetwork;
```
matteojug, line 40: Why do we need these 4 lines? We only care about mainnet and regtest, so the `addressNetwork = isMainnet ? Network.mainnet : Network.regtest;` above should cover all we need.
Then if you supply a testnet address it would fail validation, which is fine as we don't work on testnet.

src/util/decode-bitcoin-address.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { bech32 } = addressInfo;
+  let hash: Uint8Array;
+  if (bech32) {
+    hash = bitcoin.address.fromBech32(address).data;
+  } else {
+    hash = bitcoin.address.fromBase58Check(address).hash;
+  }
+
+  return {
+    type: typeMapping[addressInfo.type],
```
matteojug, line 23: I'm wondering if we should ensure `addressInfo.type` exists in `typeMapping`: right now that is always the case, but if `bitcoin-address-validation` were to update and introduce some other `addresstype`, this would return undefined and maybe bad things could happen downstream

src/comps/temp-withdraw.tsx
hozzjss: the validation lib defaults to testnet when using base58 addresses the same address is valid for regtest, as explained in the comments, so it would break if the passed address is regtest base58 and the network is regtest

src/util/decode-bitcoin-address.ts
hozzjss: that is covered since the package is pinned to the exact version that supports all these types atm and will always do since the version is pinned

src/comps/temp-withdraw.tsx
matteojug: Ah got it. Can you expand a bit the comment to make it clear what's happening? Basically as you did in this convo

src/util/decode-bitcoin-address.ts
matteojug: It is, right now, but if we update the dependency there's nothing that will flag if there's a new address type we don't support.

hozzjss: Fair enough will throw an error if the returned type is incorrect

matteojug (approved):

PR #113, opened 2025-03-05
add hello world on an empty reskin page

hozzjss:
This separates the old bridge interface in root `/` and the reskin`/` which is now just text with no relation to the old bridge ui

https://github.com/user-attachments/assets/a9a657ea-c3bb-4945-b0fa-a08d60e9b72b

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -3,36 +3,21 @@
 @tailwind utilities;
 
 :root {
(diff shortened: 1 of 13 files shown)


aldur (commented):
Looks like a good start to my limited frontend skills :)

Leaving it to @matteojug for a proper review.

PR #114, opened 2025-03-05
Chore/reskin theme

hozzjss:
This adds switcher between light and dark mode and adds a modifier for that

https://github.com/user-attachments/assets/3fead324-8908-4276-a2e6-d18da4df5155

### DIFF
--- a/src/app/reskin/components/header.tsx
+++ b/src/app/reskin/components/header.tsx
@@ -1,0 +1,21 @@
+"use client";
+
+import { themeAtom } from "@/util/atoms";
+import { useAtom } from "jotai";
+
+export default function Header() {
+  const [theme, setTheme] = useAtom(themeAtom);
+  return (
+    <header>
+      <div className="mx-auto max-w-7xl px-4 py-6 sm:px-6 lg:px-8 flex justify-between">
+        <h1 className="text-3xl font-bold tracking-tight">SBTC Bridge</h1>
+        <button
+          className="bg-blue-500 hover:bg-blue-700 text-white font-bold py-2 px-4 rounded-sm"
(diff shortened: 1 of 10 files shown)


src/util/atoms.ts
```diff
@@ -61,3 +61,10 @@ export const walletInfoAtom = atomWithStorage<{
     musig: null,
   },
 });
+
+export const themeAtom = atomWithStorage<"light" | "dark">(
+  "themeV1",
+  globalThis.window && window.matchMedia("(prefers-color-scheme: dark)").matches
+    ? "dark"
+    : "light",
```
matteojug, line 69: Do we want to default to light mode if no preferences? Given the mocks were mostly for dark mode.

src/app/reskin/components/header.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { themeAtom } from "@/util/atoms";
+import { useAtom } from "jotai";
+
+export default function Header() {
+  const [theme, setTheme] = useAtom(themeAtom);
+  return (
+    <header>
+      <div className="mx-auto max-w-7xl px-4 py-6 sm:px-6 lg:px-8 flex justify-between">
+        <h1 className="text-3xl font-bold tracking-tight">SBTC Bridge</h1>
```
matteojug, line 11: not sure if this comes from some mock, but usually we use `sBTC`

matteojug (approved):

src/util/atoms.ts
hozzjss: good question, not sure of the answer but will probably be answered down the process

src/app/reskin/components/header.tsx
hozzjss: this is still WIP actual header is next

PR #119, opened 2025-03-09
Deposit timeline component

hozzjss:
This adds the timeline part of deposits ui, it should not show up in mobile

https://github.com/user-attachments/assets/5433ac3f-5f85-43e1-9a05-e3c770a0886c


src/app/reskin/components/stepper/deposit-timeline.tsx
```diff
@@ -0,0 +1,70 @@
...
+      sBTC will be sent to a STX address. Connecting a wallet will auto-fill
+      this in, but feel free to submit another address.
+    </span>
+  );
+}
+
+const { useStepper, utils } = defineStepper(
+  {
+    id: "amount",
+    title: "Select Withdrawal Amount",
```
matteojug, line 21: nit: deposit amount

src/app/reskin/components/stepper/deposit-timeline.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+function DepositTimeline() {
+  const stepper = useStepper();
+
+  const isDesktop = useMediaQuery("(min-width: 1024px)");
+  const currentIndex = utils.getIndex(stepper.current.id);
+
+  return (
+    isDesktop && (
+      <div className="pt-6 pb-10 px-8 border rounded-2xl border-black border-opacity-20 dark:border-white dark:border-opacity-20 w-96">
```
matteojug, line 44: nit: do we need `dark:border-opacity-20` given we already have `border-opacity-20`?

src/app/reskin/components/stepper/deposit/amount.tsx
```diff
@@ -0,0 +1,22 @@
...
+      const isUnlimited = response.perDepositMinimum === 0;
+      // to handle devenv situations
+      return isUnlimited ? 0.0001 : response.perDepositMinimum / 1e8;
+    },
+  });
+  return (
+    minimumDeposit && (
+      <span className="opacity-60">
+        How much BTC are you transferring over to sBTC? Enter an amount that’s
+        above the minimum ({minimumDeposit} BTC)
```
matteojug, line 18: Should we round `minimumDeposit`? To avoid displaying strange numbers (eg a lot of decimals, or scientific notation).

src/app/reskin/components/stepper/deposit/amount.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { useQuery } from "@tanstack/react-query";
+
+export function AmountDescription({}) {
+  const { data: minimumDeposit } = useQuery({
+    queryKey: ["emilyLimits"],
+    queryFn: async () => {
+      const response = await getEmilyLimits();
+      const isUnlimited = response.perDepositMinimum === 0;
+      // to handle devenv situations
+      return isUnlimited ? 0.0001 : response.perDepositMinimum / 1e8;
```
matteojug, line 11: Why the `0.0001` instead of `0` if we don't have a limit?

src/app/reskin/components/stepper/deposit/status.tsx
```diff
@@ -0,0 +1,79 @@
...
+  return (
+    <div className="flex flex-col items-center justify-center mt-4">
+      <LoadingIndicator />
+      <div className="h-8 flex items-center text-xl leading-normal tracking-[-0.02em] text-center text-reskin-dark-gray dark:text-midGray uppercase mt-2">
+        Processing
+      </div>
+      <div>
+        ( Estimation{" "}
+        <HyperLink href="https://mempool.space/tx/4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b">
+          5min
```
matteojug, line 34: qq: do we want the estimation to be the actual one (eg, from mempool based on the fee) or just a static text?

src/app/reskin/components/stepper/deposit/amount.tsx
hozzjss: Good point I think we could fix it later after feedback assuming that the minimum stays at 0.001 will be okay

src/app/reskin/components/stepper/deposit/amount.tsx
hozzjss: arbitrary value can change it to anything assuming that we use that for testing in devnet and devenv

src/app/reskin/components/stepper/deposit/status.tsx
hozzjss: we can just use static text 10 minutes I'm thinking this might be something we change later but for now MVP wise just static text might be good

src/app/reskin/components/stepper/deposit/amount.tsx
matteojug: Why not just zero if it's zero?

hozzjss: because 0 btc is not valid amount to send 1 sat is

hozzjss: I set it to 1 sat

matteojug (approved):

src/app/reskin/components/stepper/deposit/amount.tsx
matteojug: Not a blocker, but here we talk about the min wrt emily, so zero it's a valid number. If we are talking about valid bitcoin amount, then we should use a dust value.

PR #120, opened 2025-03-11
add feature flags

hozzjss:
This is a simple PR just to make sure when pushing the v1 withdrawals the reskin isn't shipped with it. 

introducing a new env var
```sh
FEATURE_FLAGS=
```
should be a comma separated values list with the enabled features this only encompasses new withdrawals and reskin features in the withdrawals release at the 26th we'll need it to be set to 
```sh
FEATURE_FLAGS=withdrawals
```

@fabergat


hozzjss:
@matteojug mentioning requesting a review friend!

matteojug (approved):

PR #122, opened 2025-03-12
Make the set connected btc or stx addresses optional

hozzjss:
This also enforces having at least one but not the other to mark the user connected This would mean that consumers of that state request it explicitly if it's not found

### DIFF
--- a/src/comps/ConnectWallet.tsx
+++ b/src/comps/ConnectWallet.tsx
@@ -21,6 +21,7 @@ import {
   getAddressesFordefi,
 } from "@/util/wallet-utils/src/getAddress";
 import { useAsignaConnect } from "@asigna/btc-connect";
+import { AddressVersion, createAddress } from "@stacks/transactions";
 
 const WALLET_PROVIDERS = [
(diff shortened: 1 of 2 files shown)


src/comps/ConnectWallet.tsx
```diff
@@ -49,6 +50,25 @@ const WALLET_PROVIDERS = [
   },
 ];
 
+const isMainnetWallet = (addresses: Awaited<ReturnType<getAddresses>>) => {
+  const isBitcoinMainnetAddress =
+    addresses.payment?.address?.startsWith("bc1") ||
+    addresses.payment?.address?.startsWith("3");
```
matteojug, line 56: qq: why don't we support P2PKH addresses?

matteojug (dismissed):

src/comps/ConnectWallet.tsx
hozzjss: Good idea let's incorporate it!

matteojug (approved):

PR #126, opened 2025-03-13
Chore/request only required connection

hozzjss:
fix copy for deposit address, and only request the required address

### DIFF
--- a/src/comps/core/Form.tsx
+++ b/src/comps/core/Form.tsx
@@ -53,9 +53,8 @@ export const FlowForm = ({
     if (requiredConnection === "stx" && !isSTXConnected) {
       return "Connect Stacks wallet";
     }
-    if (!isBTCConnected || !isSTXConnected) {
-      return "Connect wallet";
-    }
+
+    return "Connect wallet";
   }, [requiredConnection, walletInfo.addresses]);
   const setShowConnectWallet = useSetAtom(showConnectWalletAtom);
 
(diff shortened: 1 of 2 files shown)


src/comps/core/Form.tsx
```diff
@@ -14,6 +14,7 @@ type FlowFormProps = {
   children?: React.ReactNode;
   validationSchema?: YupSchema;
   disabled?: boolean;
+  requiredConnection?: "stx" | "btc" | "both";
```
matteojug, line 17: qq: do we expect to have a flow requiring both at some time, or this is just a _why not_ thing?

src/comps/core/Form.tsx
```diff
@@ -26,9 +27,36 @@ export const FlowForm = ({
...
+    const isSTXConnected = !!stacks?.address;
+    if (requiredConnection === "btc" && !isBTCConnected) {
+      return "Connect Bitcoin wallet";
+    }
+    if (requiredConnection === "stx" && !isSTXConnected) {
+      return "Connect Stacks wallet";
+    }
+    if (!isBTCConnected || !isSTXConnected) {
+      return "Connect wallet";
+    }
```
matteojug, line 58: nit: if (eg) `requiredConnection === "stx" && isSTXConnected`, this would return `Connect wallet`. It's fine since the button is guarded by `isConnected`, but for readability we could just return `Connect wallet` (removing this last if condition).

matteojug (approved):

src/comps/core/Form.tsx
hozzjss: a whynot thing

matteojug (approved):

PR #127, opened 2025-03-13
Add ability to connect asigna stx multisig

hozzjss:
This only get the stacks address the rest of the implementation would only support withdrawals

### DIFF
--- a/src/util/wallet-utils/src/getAddress.ts
+++ b/src/util/wallet-utils/src/getAddress.ts
@@ -106,7 +106,7 @@ export const getAddressesFordefi: getAddresses = async (params) => {
 };
 
 export const getAddressesAsigna: getAddresses = async (params) => {
-  return Promise.race([getBTCAddressAsigna(params), getStxAddressAsigna()]);
+  return Promise.any([getBTCAddressAsigna(params), getStxAddressAsigna()]);
 };
 
 const getBTCAddressAsigna: getAddresses = async (params) => {
--- a/src/util/wallet-utils/src/getAddress.ts
+++ b/src/util/wallet-utils/src/getAddress.ts
@@ -145,6 +145,10 @@ const getStxAddressAsigna: getAddresses = async () => {
       AsignaIframeProvider as any,
     );
   });
+  // the thing is and don't tell anybody asigna sets mainnet and testnet address to the same address
+  // if it's testnet then it sets both of them to testnet
+  // if it's mainnet it sets both of them to mainnet
+  // musigs aren't portable through networks
   const stxAddress = result.userSession.loadUserData().profile.stxAddress
     .mainnet as string;
   return {
--- a/src/util/wallet-utils/src/getAddress.ts
+++ b/src/util/wallet-utils/src/getAddress.ts
@@ -1,11 +1,18 @@
 import { AddressPurpose, request, RpcSuccessResponse } from "sats-connect";
 import { DefaultNetworkConfigurations } from "@leather.io/models";
 
+import {
+  authenticate,
+  AppConfig,
+  UserSession,
+  FinishedAuthData,
+} from "@stacks/connect";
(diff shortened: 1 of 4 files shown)


src/util/wallet-utils/src/util/asigna-provider.ts
matteojug, line 1: nit: can you add a comment at the top saying this comes from https://github.com/hirosystems/connect/blob/main/packages/connect/src/asigna.ts? So that if something breaks we know where to look.

Also, to confirm, we cannot rely on the exported `initializeAsignaProvider` because of `document.referrer` being unreliable, correct? If so, can you also add a mention to this? Again, to avoid someone replacing it and then wasting time rediscovering why it wasn't used 😅

src/util/wallet-utils/src/getAddress.ts
```diff
@@ -99,10 +106,15 @@ export const getAddressesFordefi: getAddresses = async (params) => {
 };
 
 export const getAddressesAsigna: getAddresses = async (params) => {
+  return Promise.race([getBTCAddressAsigna(params), getStxAddressAsigna()]);
```
matteojug, line 109: qq: what happens calling `params.action()` (ie, `connect` from `useAsignaConnect` from `@asigna/btc-connect`) when using stx-asigna?

src/util/wallet-utils/src/getAddress.ts
```diff
@@ -116,6 +128,35 @@ export const getAddressesAsigna: getAddresses = async (params) => {
...
+          icon: window.location.origin + "/icon.png",
+        },
+        onFinish: res,
+        onCancel: rej,
+      },
+      AsignaIframeProvider as any,
+    );
+  });
+  const stxAddress = result.userSession.loadUserData().profile.stxAddress
+    .mainnet as string;
```
matteojug, line 149: What happens if the session data only has testnet address? I guess it will not happen (right now) with asigna, but maybe we should throw an error just for future-proofing

src/util/wallet-utils/src/getAddress.ts
matteojug, line 109: IIUC `race` will resolve with the first resolution, being it a success or failure. Is this what we want? vs using `any`

hozzjss: it's ignored since getStxAddress asigna wins the race

src/util/wallet-utils/src/getAddress.ts
hozzjss: the thing is and don't tell anybody asigna sets mainnet and testnet address to the same address, if it's testnet then it sets both of them to testnet if it's mainnet it sets both of them to mainnet musigs aren't portable through networks

src/util/wallet-utils/src/getAddress.ts
hozzjss: good idea asigna might change their apis in the future any would future proof this

src/util/wallet-utils/src/getAddress.ts
hozzjss: added a comment explaining this as well

src/util/wallet-utils/src/getAddress.ts
matteojug: Yeah, but assuming we are not racing, what happens to that call? It stalls forever, return an error or wreak havoc?

hozzjss: Good q it will be basically waiting in the bg but nothing to worry about here memory leak wise according to [this guy](https://stackoverflow.com/posts/61136891/revisions) at least lol

matteojug: Got it, thanks. We should double check what happens now that we are using `any` (I'm thinking about a scenario where the btc asigna stalls while the stx asigna fails), but I guess it's fine for now.

PR #130, opened 2025-03-13
Feat/withdraw flow

hozzjss:
This introduces the withdraw status page this watches the status of the withdrawal


src/actions/get-withdrawal-data.ts
```diff
@@ -0,0 +1,73 @@
...
+  hex: string;
+  repr: string;
+  name: string;
+  type: string;
+};
+
+export async function getWithdrawalInfo(txid: string) {
+  const tx = await fetch(`${env.STACKS_API_URL}/extended/v1/tx/${txid}`);
+  const txData = await tx.json();
+  const contractCall = txData.contract_call as contractCallData;
```
matteojug, line 27: Should we handle the not found case explicitly? As in, if the tx is not yet available in the stack api we are querying.

Also below for emily returning a non success.

src/actions/get-withdrawal-data.ts
```diff
@@ -0,0 +1,73 @@
...
+  ).value;
+
+  const address = encodeBitcoinAddress(
+    recipient.hashbytes.value,
+    recipient.version.value,
+    getBitcoinNetwork(env.WALLET_NETWORK),
+  );
+  let data = {
+    address,
+    amount: Number(amount),
```
matteojug, line 46: I guess this conversion is safe as bitcoin max money in sats is less than MAX_SAFE_INTEGER, but why not returning the big int directly?

src/actions/get-withdrawal-data.ts
```diff
@@ -0,0 +1,73 @@
...
+    const requestId = printEventDeserialized.value["request-id"]
+      .value as bigint;
+
+    const withdrawal = await fetch(`${env.EMILY_URL}/withdrawal/${requestId}`);
+
+    const withdrawalData = await withdrawal.json();
+    status = WithdrawalStatus[withdrawalData.status as WithdrawalStatus];
+  }
+
+  if (txData.tx_status !== "pending" && txData.tx_status !== "success") {
```
matteojug, line 66: nit: else if to remove the `txData.tx_status !== "success"` bit

src/app/withdraw/[txid]/components/withdrawal-stepper.tsx
```diff
@@ -0,0 +1,101 @@
...
+}: {
+  name: string;
+  id: (typeof steps)[number]["id"];
+  stepper: ReturnType<typeof useStepper>;
+}) {
+  const myIndex = utils.getIndex(id);
+
+  const isCurrentStep = stepper.current.id === id;
+  const currentIndex = utils.getIndex(stepper.current.id);
+  const isCompleted = myIndex <= currentIndex;
```
Comment (matteojug) [inline src/app/withdraw/[txid]/components/withdrawal-stepper.tsx:35]:
```suggestion
  const isCompleted = myIndex < currentIndex;
```

src/app/withdraw/[txid]/page.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { getWithdrawalInfo } from "@/actions/get-withdrawal-data";
+
+export default async function WithdrawalStatusPage({
+  params,
+}: {
+  params: Promise<{ txid: string }>;
+}) {
+  const { txid } = await params;
+
+  const result = await getWithdrawalInfo(txid);
```
Comment (matteojug) [inline src/app/withdraw/[txid]/page.tsx:13]: What's the benefit of passing `result` to `TrackWithdrawalStatus`, given we use a `useQuery` there? vs just passing `txid` to `TrackWithdrawalStatus`

src/app/withdraw/components/withdraw-client.tsx
```diff
@@ -39,38 +48,72 @@ const decodeBitcoinAddressToClarityRecipient = (
...
 
-const BasicWithdraw = () => {
-  const [txId, setTxId] = useState<string | null>(null);
+const Withdraw = () => {
   const { addresses, selectedWallet } = useAtomValue(walletInfoAtom);
+  const { data: satsBalance } = useSBTCBalance({
+    address: addresses.stacks?.address,
+  });
+
+  // const stxAddress = addresses.stacks?.address;
```
matteojug, line 78: nit: leftover

src/app/withdraw/components/withdraw-client.tsx
```diff
@@ -145,49 +186,70 @@ const BasicWithdraw = () => {
...
+              setFieldValue("address", value);
+              stepper.next();
+            }}
+            stepper={stepper as Stepper<any>}
+          />
+        ),
+        confirm: () => (
+          <WithdrawConfirm
+            amount={Number(values.amount)}
+            btcAddress={values.address || ""}
```
matteojug, line 231: qq: why the `|| ""` here?

src/util/constants.ts
```diff
@@ -0,0 +1 @@
+export const MAX_WITHDRAWAL_FEE = 0.0008 * 1e8;
```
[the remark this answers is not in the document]
hozzjss: Readability's sake computationally this is negligible but I'm letting devs know that you should think in terms of BTC not sats until it comes to server requests or business logic

src/app/withdraw/components/withdraw-confirm.tsx
```diff
@@ -0,0 +1,51 @@
...
+        <div className="flex flex-col gap-1">
+          <SubText>Amount selected to Withdraw</SubText>
+          <p className="text-black font-Matter font-semibold text-sm">
+            {amount} sBTC
+          </p>
+        </div>
+        <div className="flex flex-col gap-1">
+          <SubText>Bitcoin address to withdraw to</SubText>
+          <p className="text-black font-Matter font-semibold text-sm">
+            {useShortAddress(btcAddress)}
```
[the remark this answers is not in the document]
hozzjss: makes sense the only problem is that I need text wrap for long addresses on mobile I'm thinking maybe I show the full address on hover and give the user the ability to copy their address maybe wdyt?

src/util/yup/test-stx-address.ts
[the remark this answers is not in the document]
hozzjss: makes sense

src/actions/get-withdrawal-data.ts
```diff
@@ -0,0 +1,73 @@
+"use server";
```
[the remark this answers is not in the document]
hozzjss: Good question, the idea is to obfuscate the stacks api url, and we'll need a special api key for sure from them

src/actions/get-withdrawal-data.ts
hozzjss: the tx not yet available in the stacks api is not possible since we broadcast the tx ourselves, if emily returns a non success it will be retried

src/actions/get-withdrawal-data.ts
hozzjss: consumers are components that don't play with bigints

src/app/withdraw/[txid]/page.tsx
Comment (hozzjss) [reply src/app/withdraw/[txid]/page.tsx:13]: good question so this is so that the query has initial data indicating whether it should be ran or not basically disabling the query at init and giving the component data on render

src/app/withdraw/components/withdraw-client.tsx
hozzjss: good catch

src/app/withdraw/components/withdraw-confirm.tsx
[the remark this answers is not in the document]
hozzjss: what do you think?
https://github.com/user-attachments/assets/6a72a2e4-0d48-40a6-9cb3-dafe415ea632

src/util/constants.ts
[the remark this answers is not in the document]
matteojug: That makes sense, but since we want to use an integer value maybe it's better to explicitly use an int value, so that we don't have to reason about possible float rounding artifact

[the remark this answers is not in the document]
hozzjss: makes sense okay

src/app/withdraw/components/withdraw-confirm.tsx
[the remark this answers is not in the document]
matteojug: Nice! Maybe it's not clear in terms of UX but we can address that later, as long as there's a way to see the full address it's fine

src/actions/get-withdrawal-data.ts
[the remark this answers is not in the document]
matteojug: Can you open an issue to remember to get (and use) an API token for server side calls to hiro apis?

src/actions/get-withdrawal-data.ts
matteojug: Just to confirm: if emily returns an error `withdrawal.json()` will throw, or if it's a json encoded error the returned `status` may be undefined. Is this handled correctly?

src/hooks/use-short-address.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const copyOnClick = () => {
+    navigator.clipboard.writeText(address);
+    notify({
+      message: "Address copied to clipboard",
+      type: NotificationStatusType.SUCCESS,
+    });
+  };
+
+  const shortenedAddress = useMemo(() => {
+    if (address.length > 10) {
```
matteojug, line 20: nit: it shouldn't matter given the nature of the addresses, but still
```suggestion
    if (address.length > 20) {
```

src/actions/get-withdrawal-data.ts
hozzjss: no good catch need to handle that correctly however if the request returns ok status I'm assuming must exist

hozzjss: should be fixed now

src/actions/get-withdrawal-data.ts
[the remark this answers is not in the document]
hozzjss: done you might need to add appropriate labels details there https://github.com/stacks-network/sbtc/issues/1517
I don't have my old stacks org privileges no mo'

matteojug (approved):

PR #133, opened 2025-03-17
update @stacks/transactions and add asigna contract call

hozzjss:
Asigna integration for withdrawals and that marks all supported wallets as integrated!

### DIFF
--- a/src/util/wallet-utils/src/callContract.ts
+++ b/src/util/wallet-utils/src/callContract.ts
@@ -5,6 +5,14 @@ import {
   getFordefiBTCProviderOrThrow,
   getLeatherBTCProviderOrThrow,
 } from "./util/btc-provider";
+import { openContractCall } from "@stacks/connect";
+import { AsignaIframeProvider } from "./util/asigna-provider";
+import {
+  addressToString,
+  ContractCallPayload,
+  deserializeTransaction,
+  wireToPostCondition,
+} from "@stacks/transactions";
 
 type RequestArgs = {
   txHex: string;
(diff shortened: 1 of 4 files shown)


src/util/wallet-utils/src/callContract.ts
```diff
@@ -47,5 +55,25 @@ export const callContractFordefi: CallContract = async (params) => {
 };
 
 export const callContractAsigna: CallContract = async (params) => {
-  throw new Error("Not implemented");
+  const tx = deserializeTransaction(params.txHex);
+  const payload = tx.payload as ContractCallPayload;
+
+  return new Promise<string>((resolve, reject) =>
+    openContractCall(
```
matteojug, line 62: If I'm reading the docs correctly, this will alwsa submit the tx (https://www.npmjs.com/package/@stacks/connect/v/7.10.1-alpha.f3096a1.0#calling-smart-contracts-opencontractcall):
> Calling Smart-Contracts (openContractCall)
> Calling smart-contracts lets users interact with the blockchain through transactions.
>
> The snippet below will open the wallet to confirm and broadcast a smart-contract transaction.

hozzjss: matteo you're a genius

matteojug: Oh right, then I guess we need to change `src/app/withdraw/components/withdraw-client.tsx` to skip the broadcast for asigna, as well as considering the case of tx not existing in `getWithdrawalInfo`

hozzjss: exactly my thinking

hozzjss: so unfortunately interaction with the wallet atm is very limited will keep working on it but for the time being the minimum is met i think which is get the tx out starting the process, the idea tho is in the status, which can be polled by using the initiate tx id /withdraw/[txid] we need to somehow to go from withdraw -> asigna sign and execute -> status the flow breaks in the middle step

PR #135, opened 2025-03-18
less dependence on hiro api to get withdrawal status

hozzjss:
basically I would only need the hiro stacks api to check the tx status and to get the request id, the route would work with both tx id and request id but would be faster with request id since that would be only one request to emily instead of one to the stacks api and another to emily

### DIFF
--- a/src/actions/get-withdrawal-data.ts
+++ b/src/actions/get-withdrawal-data.ts
@@ -10,6 +10,7 @@ import getBitcoinNetwork from "@/util/get-bitcoin-network";
 import { hiroClient } from "./hiro-fetch";
 
 import { getEmilyWithdrawal } from "./get-emily-withdrawal";
+import { getRegistryWithdrawal } from "./get-registry-withdrawal";
 
 type contractCallData = {
   contract_id: string;
@@ -34,7 +35,16 @@ export async function getWithdrawalInfo(txidOrRequestId: string): Promise<{
   const isRequestId = txidOrRequestId.length < 64;
 
   if (isRequestId) {
-    return await getEmilyWithdrawal(txidOrRequestId);
+    try {
+      return getEmilyWithdrawal(txidOrRequestId);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "error fetching emily withdrawal falling back to registry",
+        // e,
+      );
+      return getRegistryWithdrawal(txidOrRequestId);
+    }
   }
 
   const tx = await hiroClient.fetch(
--- a/src/actions/get-withdrawal-data.ts
+++ b/src/actions/get-withdrawal-data.ts
@@ -31,6 +31,8 @@ export async function getWithdrawalInfo(txidOrRequestId: string): Promise<{
   address: string;
   requestId: string | null;
   amount: number;
+  stacksTx?: string;
+  bitcoinTx?: string;
 }> {
   const isRequestId = txidOrRequestId.length < 64;
 
--- a/src/actions/get-withdrawal-data.ts
+++ b/src/actions/get-withdrawal-data.ts
@@ -6,7 +6,10 @@ import { BufferCV, Cl, TupleCV, UIntCV } from "@stacks/transactions";
 import { encodeBitcoinAddress } from "@/util/decode-bitcoin-address";
 import { WithdrawalStatus } from "@/app/withdraw/[txid]/components/util";
 import getBitcoinNetwork from "@/util/get-bitcoin-network";
+
 import { hiroClient } from "./hiro-fetch";
+
+import { getEmilyWithdrawal } from "./get-emily-withdrawal";
 
 type contractCallData = {
   contract_id: string;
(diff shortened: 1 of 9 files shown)


src/actions/get-withdrawal-data.ts
```diff
@@ -57,13 +72,7 @@ export async function getWithdrawalInfo(txid: string) {
...
       .value as bigint;
 
-    const withdrawal = await fetch(`${env.EMILY_URL}/withdrawal/${requestId}`);
-
-    if (!withdrawal.ok) {
-      throw new Error(`Failed to fetch withdrawal data: ${withdrawal.status}`);
-    }
-    const withdrawalData = await withdrawal.json();
-    status = WithdrawalStatus[withdrawalData.status as WithdrawalStatus];
+    return await getEmilyWithdrawal(String(requestId));
```
matteojug, line 75: We may want to suppress the errors from this one: there may be a time window in which the node we ask for the tx return success but Emily is still missing the withdraw. Ideally we also want to avoid redirecting until this is successful.

matteojug: Oh it may already be the case:
 - We have initial data without the requestId, so we would not redirect
 - If this fails, we return an error so the caller doesn't get a requestId

I guess the only possible issue is if the initial call to populate initialData ends up in the scenario above (or in general some fetch fails), thus returning an error: what happens in `WithdrawalStatusPage` if `getWithdrawalInfo` throws?

hozzjss: Application error: a server-side exception has occurred (see the server logs for more information).
Digest: 3691147983

😭

hozzjss: so I thought about this, and this scenario can't happen because the request id is never returned until `getEmilyWithdrawal` returns it, basically if for example if I start a withdrawal and shut down emily api the request will keep returning pending or break and it will keep retrying, the case we need to handle here is if emily is down and the url is a request id in this case we could have a simple 500/404 page that retries fetching indefinitely

src/app/withdraw/[txid]/components/withdrawal-status.tsx
```diff
@@ -113,20 +136,31 @@ function Content(initialData: Props) {
...
+              getStacksNetwork(bridgeConfig.WALLET_NETWORK),
+            )}
             target="_blank"
             rel="noreferrer"
           >
             View stacks tx
           </a>
-        </div>
-      )}
+        )}
```
Comment (matteojug) [inline src/app/withdraw/[txid]/components/withdrawal-status.tsx:152]: We should probably keep a button to link to the submitted tx (the init withdrawal) -- either a separate one, or reusing this one if `stacksTx` is not set. This current one, coming from `fulfillment`, is the stacks txid of the withdrawal accept/reject, so you only have it after the request is completed.

src/lib/get-explorer-url.ts
```diff
@@ -0,0 +1,7 @@
+export const getExplorerUrl = (
```
matteojug, line 1: Nice! Can we also use this in deposit and footer? Either here on in a different PR

hozzjss: I can do it here

src/app/withdraw/[txid]/components/withdrawal-status.tsx
Comment (hozzjss) [reply src/app/withdraw/[txid]/components/withdrawal-status.tsx:152]: fair I can keep it but the problem here is that i need it from the emily api after the switch to fetching the request from emily

src/actions/get-withdrawal-data.ts
```diff
@@ -22,9 +26,31 @@ type FunctionArg = {
...
+  if (isRequestId) {
+    try {
+      return getEmilyWithdrawal(txidOrRequestId);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        "error fetching emily withdrawal falling back to registry",
+        // e,
+      );
+      return getRegistryWithdrawal(txidOrRequestId);
```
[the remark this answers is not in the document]
hozzjss: fair nuff

src/app/withdraw/[txid]/components/withdrawal-status.tsx
```diff
@@ -63,6 +72,20 @@ function Content(initialData: Props) {
...
+
+  useEffect(() => {
+    // if the route is request id format do nothing
+    if (txid.length < 64) {
+      return;
+    }
+
+    // if it is a tx id and we have a request id navigate to request id
+    if (requestId) {
+      router.push(`/withdraw/${requestId}`);
```
Comment (matteojug) [inline src/app/withdraw/[txid]/components/withdrawal-status.tsx:86]: mmm there's a nuisance with this: the request id does depend on the init withdrawal submission order, and on reorgs you cannot assume it's going to play in the same order (even with tx replay), so I think we shouldn't redirect to request id until the request id "is final"; on the signers we assume 6 bitcoin blocks to be final enough (it's not related to the signers, just suggesting a number).

We can address this separately from here, just flagging it

Comment (hozzjss) [reply src/app/withdraw/[txid]/components/withdrawal-status.tsx:86]: so the withdrawal id is never 100% deterministic until enough finality is reached on bitcoin 6 blocks is the minimum

src/app/withdraw/[txid]/components/withdrawal-status.tsx
Comment (hozzjss) [reply src/app/withdraw/[txid]/components/withdrawal-status.tsx:152]: and also we might need the reject tx, basically accepted withdrawals get a special treatment

matteojug (approved):

PR #137, opened 2025-03-18
add hiro api key and use it for all operations in the bridge

hozzjss:


src/actions/hiro-fetch.ts
```diff
@@ -0,0 +1,17 @@
+import { env } from "@/env";
```
matteojug, line 1: Do we need `"use server";` here?

matteojug (approved):

matteojug (approved):

PR #138, opened 2025-03-18
disable deposits if aggregate key is not available

hozzjss:
resolves https://github.com/stacks-network/sbtc/issues/1393

### DIFF
--- a/src/actions/get-aggregate-key.ts
+++ b/src/actions/get-aggregate-key.ts
@@ -19,7 +19,7 @@ export async function getAggregateKey() {
     },
   })) as BufferCV;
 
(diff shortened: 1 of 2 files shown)


src/actions/get-aggregate-key.ts
```diff
@@ -19,5 +19,9 @@ export async function getAggregateKey() {
     },
   })) as BufferCV;
 
+  if (result.value.length < 64) {
```
matteojug, line 22: Shouldn't this be 66?

hozzjss: makes sense

PR #140, opened 2025-03-24
fix sbtc balance not loading and footer aggregate address not working

hozzjss:
This basically delays requesting the balance from the action until the ui is loaded, and for footer uses full SSR to make sure the env is passed accurately

### DIFF
--- a/src/comps/footer.tsx
+++ b/src/comps/footer.tsx
@@ -1,6 +1,6 @@
-"use client";
+"use server";
 import { getAggregateKey } from "@/actions/get-aggregate-key";
-import { useQuery } from "@tanstack/react-query";
+
 import { payments } from "bitcoinjs-lib";
 import Image from "next/image";
 import { hexToBytes } from "@stacks/common";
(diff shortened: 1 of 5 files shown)


src/comps/footer.tsx
```diff
@@ -32,21 +32,17 @@ const contracts = [
...
 
-export default function Footer({ config }: { config: BridgeConfig }) {
+export default async function Footer({ config }: { config: BridgeConfig }) {
   const network = getBitcoinNetwork(config.WALLET_NETWORK);
   const { SUPPORT_LINK: supportLink } = config;
-  const { data: aggregateAddress } = useQuery({
-    queryKey: ["aggregateAddress"],
-    queryFn: async () => {
-      const response = await getAggregateKey();
+  const response = await getAggregateKey();
```
matteojug, line 38: if this readonly call throws, do we get a "bad" error?

hozzjss: good catch this might indeed fail reloading should fix it but it shouldn't happen much since we're depending on hiro infra and this is an internal request

hozzjss: fixed

matteojug (approved):

PR #142, opened 2025-03-24
cuter nav for bridge between deposit and withdraw

hozzjss:
check this out

https://github.com/user-attachments/assets/47b50e08-5ad5-4891-8414-4e6e1e2b460d

### DIFF
--- a/src/comps/Header.tsx
+++ b/src/comps/Header.tsx
@@ -102,11 +102,6 @@ const Header = ({ config }: { config: BridgeConfig }) => {
             {/* <h5 className="font-Matter text-xs text-black tracking-wide ">
               LEARN MORE
             </h5> */}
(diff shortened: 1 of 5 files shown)


src/comps/Header.tsx
```diff
@@ -102,11 +102,6 @@ const Header = ({ config }: { config: BridgeConfig }) => {
             {/* <h5 className="font-Matter text-xs text-black tracking-wide ">
               LEARN MORE
             </h5> */}
-            {isConnected && config.FEATURE_FLAGS.includes("withdrawals") && (
```
matteojug, line 105: since now withdrawal UI is no longer feature gated, can we also remove the withdrawal redirect gate?

hozzjss: makesh shense

hozzjss: done

PR #143, opened 2025-03-24
add actual max fee approximation to withdraw

hozzjss:
This shows the estimated thousands of sats that will be reserved as max fee it assumes that it will be higher than 1K sats instead of the fixed fee (80K)

image

### DIFF
--- a/src/app/withdraw/components/withdraw-amount.tsx
+++ b/src/app/withdraw/components/withdraw-amount.tsx
@@ -7,11 +7,15 @@ export const WithdrawFlowAmount = ({
   handleSubmit,
   maxFee,
   disabled,
+  isFailed,
+  isLoading,
 }: {
   validationSchema: Schema;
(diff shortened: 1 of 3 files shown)


src/app/withdraw/components/withdraw-amount.tsx
```diff
@@ -5,10 +5,12 @@ import { Schema } from "yup";
 export const WithdrawFlowAmount = ({
   validationSchema,
   handleSubmit,
+  maxFee,
   disabled,
 }: {
   validationSchema: Schema;
   handleSubmit: (value: any) => void;
+  maxFee: number;
```
matteojug, line 13: `maxFee` can be undefined (we lie to the compiler with `maxFee!`, but we set `disabled` to `maxFee === undefined` so it's a possibility) and in that case we would get a `NaN` in the fee computation as it's not guarded by `disabled`. Am I missing something?

PR #144, opened 2025-03-25
work around current forms to use async validation

hozzjss:
This would mark the default withdrawal to `Infinity` if the returned value is null and if not the max would be whatever number value is fetched

resolves: https://github.com/stacks-network/sbtc/issues/1533

### DIFF
--- a/src/app/withdraw/components/withdraw-client.tsx
+++ b/src/app/withdraw/components/withdraw-client.tsx
@@ -252,7 +252,7 @@ const Withdraw = () => {
               if (value > maxWithdrawalCap) {
                 return getMaxError(maxWithdrawalCap);
               }
-              setFieldValue("address", value);
+              setFieldValue("amount", value);
               stepper.next();
             }}
             disabled={maxFee === undefined || emilyLimitsLoading}
--- a/src/app/withdraw/components/withdraw-client.tsx
+++ b/src/app/withdraw/components/withdraw-client.tsx
@@ -37,6 +37,7 @@ import { WithdrawConfirm } from "./withdraw-confirm";
 import { useSBTCBalance } from "@/hooks/use-sbtc-balance";
 import { useQuery } from "@tanstack/react-query";
 import { getWithdrawalMaxFee } from "@/actions/get-withdrawal-max-fee";
+import { useEmilyLimits } from "@/hooks/use-mint-caps";
 
 const decodeBitcoinAddressToClarityRecipient = (
   address: string,
@@ -87,6 +88,16 @@ const Withdraw = () => {
     },
   });
 
+  const {
+    data: emilyLimits,
+    refetch,
+    isLoading: emilyLimitsLoading,
+  } = useEmilyLimits();
+
+  const maxWithdrawal = (emilyLimits?.perWithdrawalCap || 0) / 1e8;
+  const getMaxError = (maxWithdrawal: number) => {
+    return `Withdrawal exceeds current cap of ${maxWithdrawal.toLocaleString(undefined, { maximumFractionDigits: 8 })} BTC`;
+  };
   const amountValidationSchema = useMemo(() => {
     const btcBalance = Number(satsBalance) / 1e8;
     const fee = maxFee! / 1e8;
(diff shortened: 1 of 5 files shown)


src/hooks/use-mint-caps.ts
```diff
@@ -3,15 +3,19 @@ import getEmilyLimits from "@/actions/get-emily-limits";
 import { useQuery } from "@tanstack/react-query";
 import { useCallback, useMemo } from "react";
 
+export function useEmilyLimits() {
+  return useQuery({
+    queryKey: ["deposit-max-fee"],
```
matteojug, line 8: nit: update the key? even if doesn't matter as we are using the same `useEmilyLimits` around

src/comps/core/Form.tsx
```diff
@@ -62,8 +64,16 @@ export const FlowForm = ({
...
+    onSubmit: async (values) => {
+      const error = await handleSubmit(values[nameKey]);
+      try {
+        if (error && typeof error === "string") {
+          await formik.setFieldTouched(nameKey, true, true);
+          throw new Error(error);
+        }
+      } catch (e: any) {
+        formik.setErrors({ [nameKey]: e.message });
+      }
```
matteojug, line 76: qq: what is this doing?

src/app/withdraw/components/withdraw-client.tsx
```diff
@@ -87,6 +88,16 @@ const Withdraw = () => {
     },
   });
 
+  const {
+    data: emilyLimits,
+    refetch,
+    isLoading: emilyLimitsLoading,
+  } = useEmilyLimits();
+
+  const maxWithdrawal = (emilyLimits?.perWithdrawalCap || 0) / 1e8;
```
matteojug, line 97: a null `perWithdrawalCap` is inf, also didn't we already validate this `getEmilyLimits`? Maybe we can drop the nullable

src/app/withdraw/components/withdraw-client.tsx
```diff
@@ -230,12 +242,20 @@ const Withdraw = () => {
...
+            handleSubmit={async (value) => {
+              const limits = await refetch();
+
+              const maxWithdrawalCap =
+                (limits.data?.perWithdrawalCap || 0) / 1e8;
+
+              if (value > maxWithdrawalCap) {
+                return getMaxError(maxWithdrawalCap);
+              }
+              setFieldValue("address", value);
```
matteojug, line 255: is `address` instead of `amount` correct here?

src/app/withdraw/components/withdraw-client.tsx
matteojug, line 97: Is `emilyLimits` guaranteed to have a value, even if we just called `useEmilyLimits`?

hozzjss: since this is a promise it will be nullable until it's resolved

src/app/withdraw/components/withdraw-client.tsx
hozzjss: correct

src/app/withdraw/components/withdraw-client.tsx
hozzjss: no it's not guaranteed until the action/request is fetched from the server

src/comps/core/Form.tsx
matteojug: yeah that much was clear. Why do we use `setFieldTouched`? Why throwing and then catching the error right after, instead of calling directly `setErrors`?

src/app/withdraw/components/withdraw-client.tsx
matteojug: This seems a bit strange: should we use `useMemo` or something else, to ensure we compute `maxWithdrawal` only when `emilyLimits` is ready?

src/comps/core/Form.tsx
hozzjss: good point

hozzjss: fixed

src/app/withdraw/components/withdraw-client.tsx
hozzjss: we compute this even more frequently on every render some memos aren't that important the memory fingerprint is negligible

matteojug (approved):

PR #147, opened 2025-03-25
reload sbtc balance periodically and fix active tab state for withdrawal

hozzjss:
This reloads sbtc balance after withdrawals and deposits, and fixes an issue where when the withdrawal status page is open the tab is inactive

https://github.com/user-attachments/assets/c3c8ec4b-00e8-43ec-a09c-f42acbfb6581

### DIFF
--- a/src/app/reskin/components/tabs/nav-tabs.tsx
+++ b/src/app/reskin/components/tabs/nav-tabs.tsx
@@ -11,8 +11,8 @@ export const NavTabs = ({
     <div className="fixed bottom-0 sm:sticky dark:bg-reskin-dark-gray w-full bg-white">
       <div className="flex h-20 flex-row justify-between items-center w-full sm:mx-auto max-w-5xl relative z-10">
         {tabs.map((tab, index) => {
-          const [, tabNameSpace] = tab.link.split("/");
-          const [, pathNameSpace] = pathname.split("/");
+          const [, tabNameSpace] = tab.link.replace("/reskin", "").split("/");
+          const [, pathNameSpace] = pathname.replace("/reskin", "").split("/");
 
           const isActive = pathNameSpace === tabNameSpace;
           return (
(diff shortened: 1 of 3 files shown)


src/app/withdraw/[txid]/components/withdrawal-status.tsx
```diff
@@ -86,6 +87,15 @@ function Content(initialData: Props) {
     }
   }, [stepper, status]);
 
+  useEffect(() => {
+    if (
+      status === WithdrawalStatus.confirmed ||
+      status === WithdrawalStatus.failed
```
Comment (matteojug) [inline src/app/withdraw/[txid]/components/withdrawal-status.tsx:93]: nit/fyi: we don't need the `failed` case as the "balance" we query for with `get-balance` includes both the unlocked and the locked amount, so in case of a failed withdrawal we would just unlock the locked amount (but the total will remain the same). But this is cleaner, so let's keep it as is.

src/app/reskin/components/tabs/nav-tabs.tsx
```diff
@@ -10,16 +10,22 @@ export const NavTabs = ({
...
-            key={index}
-            href={tab.link}
-            prefetch={true}
-          >
-            {tab.label}
-          </Link>
-        ))}
+        {tabs.map((tab, index) => {
+          const [, tabNameSpace] = tab.link.split("/");
+          const [, pathNameSpace] = pathname.split("/");
```
matteojug, line 15: This will work only if the first element is unique in the tab collection, but for the (current) reskin that is not the case (eg, `/reskin/withdraw` vs `/reskin`). Is it fine?

src/app/withdraw/[txid]/components/withdrawal-status.tsx
Comment (hozzjss) [reply src/app/withdraw/[txid]/components/withdrawal-status.tsx:93]: fair point

src/app/reskin/components/tabs/nav-tabs.tsx
hozzjss: fair nuff

matteojug (approved):

PR #149, opened 2025-03-26
add env var for min withdrawal amount

hozzjss:
add env var for minimum withdrawal amount


src/env.ts
```diff
@@ -28,4 +28,5 @@ export const env = {
   FEATURE_FLAGS: (process.env.FEATURE_FLAGS?.split(",") || []) as featureFlag[],
   WITHDRAWAL_FEE_MULTIPLIER: Number(process.env.WITHDRAWAL_FEE_MULTIPLIER || 4),
   HIRO_API_KEY: process.env.HIRO_API_KEY,
+  WITHDRAW_MIN_AMOUNT_SATS: Number(process.env.WITHDRAW_MIN_AMOUNT_SATS || 0),
```
matteojug, line 31: nit: we should default this to `547` instead of zero (unless we already have an explicit dust check somewhere else)

src/app/withdraw/components/withdraw-client.tsx
```diff
@@ -102,13 +102,14 @@ const Withdraw = () => {
...
   };
+  const config = useAtomValue(bridgeConfigAtom);
   const amountValidationSchema = useMemo(() => {
     const btcBalance = Number(satsBalance) / 1e8;
     const fee = maxFee! / 1e8;
     return yup.object().shape({
       amount: yup
         .number()
-        .min(0)
+        .min(config.WITHDRAW_MIN_AMOUNT_SATS)
```
matteojug, line 112: Shouldn't this be divided by `1e8` as well?

hozzjss: correct

src/env.ts
hozzjss: I think we should assume this can only be undefined in development envs making it fun to play with

matteojug (approved):

PR #150, opened 2025-03-26
use client fetch for withdrawal status

hozzjss:
Resolves: https://github.com/stacks-network/sbtc/issues/1551

### DIFF
--- a/src/app/withdraw/[txid]/components/withdrawal-status.tsx
+++ b/src/app/withdraw/[txid]/components/withdrawal-status.tsx
@@ -18,12 +18,7 @@ import { queryClient } from "@/query/client";
 
 const { useStepper, Scoped } = withdrawalStepper;
 type Props = {
-  btcAmount: number;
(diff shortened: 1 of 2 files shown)


src/app/withdraw/[txid]/components/withdrawal-status.tsx
```diff
@@ -93,66 +89,70 @@ function Content(initialData: Props) {
...
-      <div className="flex flex-col  gap-2">
-        <div className="flex flex-col gap-1">
-          <SubText>Amount selected to withdraw</SubText>
-          <p className="text-black font-Matter font-semibold text-sm">
-            {(amount / 1e8).toLocaleString(undefined, {
-              maximumFractionDigits: 8,
-            })}{" "}
-            BTC
-          </p>
+    isFetched && (
```
Comment (matteojug) [inline src/app/withdraw/[txid]/components/withdrawal-status.tsx:96]: nit: should we display something while it's not fetched?

matteojug (approved):

PR #153, opened 2025-04-01
better readme and starter env sample

hozzjss:
This PR updates the readme and the starter env to use testnet env


aldur (dismissed):

.env.sample
```diff
@@ -1,21 +1,21 @@
 BITCOIND_URL=
-EMILY_URL=
-MEMPOOL_API_URL=https://mempool.space/api
-PUBLIC_MEMPOOL_URL=https://mempool.space
+EMILY_URL=http://temp.sbtc-emily-dev.com
+MEMPOOL_API_URL=https://mempool.bitcoin.regtest.hiro.so/api
+PUBLIC_MEMPOOL_URL=https://mempool.bitcoin.regtest.hiro.so
```
matteojug, line 4: Can we use devenv instead (see https://github.com/stacks-sbtc/sbtc/blob/main/docker/docker-compose.yml#L248)? I don't think we want local runs to interact with testnet deployment.
Then we can add a paragraph on the readme about this (eg, the default env runs against sbtc devenv, go to this repo for the instructions).

.env.sample
```diff
@@ -1,21 +1,21 @@
...
 SUPPORT_LINK=
 POLLING_INTERVAL=5000
-BRIDGE_APP_URL=https://app.stacks.co
-STACKS_API_URL=https://api.hiro.so
+BRIDGE_APP_URL=http://localhost:3000
+STACKS_API_URL=https://api.testnet.hiro.so
 # comma separated values style
 # withdrawals,reskin
-FEATURE_FLAGS=
+FEATURE_FLAGS=reskin,withdrawals
```
matteojug, line 18: `withdrawals` is no longer used, also as a "sample" env why not having this empty instead?

README.md
```diff
@@ -1,39 +1,21 @@
...
 
 ```bash
-npm run dev
-# or
-yarn dev
-# or
-pnpm dev
-# or
-bun dev
+cp .env.example .env
```
matteojug, line 8: `.env.example` doesn't exists

matteojug (changes requested):

.env.sample
hozzjss: yes wanted to do this initially makes sense

.env.sample
hozzjss: makes sense

README.md
hozzjss: good catch

.env.sample
aldur: It might be worth to document feature flags in the README too, actually

.env.sample
matteojug: We recently switched to explicit ipv4 binding for devenv, can we replace `localhost` with `127.0.0.1` here as well?

README.md
```diff
@@ -2,15 +2,30 @@
...
-cp .env.example .env
+cp .env.sample .env
 yarn && yarn dev
 ```
 
 The development server will be running at [http://localhost:3000](http://localhost:3000)
 
+## Feature flags
+
+Some features will be gated by feature flags. These flags can be enabled in the .env file like so:
```
matteojug, line 24: Can we also mention that some feature flags must be set at build time, so if you are building the image with docker and you want some specific feature flags you must pass them also when building to ensure they are in effect?

.env.sample
matteojug: Sorry for the confusion, but can we move the feature flag list to the readme, and keep here only the format help (ie, line 16 and 17)?

hozzjss: done

aldur (approved):

matteojug (approved):

PR #161, opened 2025-04-22
Chore/reskin is skin

hozzjss:
this makes reskin the default skin, need to be rebased after history is merged since it's dependent i think maybe


djordon (approved):
Looks good

aldur (approved):

PR #164, opened 2025-04-28
Fix/small issues

hozzjss:
should resolve both issues where withdraw keeps fetching a non existing tx and withdraw or deposit confirm button not being disabled when submission is in progress and show a more accurate time for confirming deposits or withdrawals

### DIFF
--- a/src/actions/hiro-fetch.ts
+++ b/src/actions/hiro-fetch.ts
@@ -10,6 +10,7 @@ async function hiroFetch(url: string, options: RequestInit = {}) {
   return fetch(url, {
     ...options,
     headers: { ...options.headers, ...apiHeaders },
+    cache: "no-store",
   });
 }
 
--- a/src/app/reskin/components/forms/deposit/deposit-form.tsx
+++ b/src/app/reskin/components/forms/deposit/deposit-form.tsx
@@ -143,7 +143,15 @@ export const DepositForm = () => {
         }
       }}
     >
-      {({ errors, touched, isValid, values, submitForm, validateForm }) => (
+      {({
+        errors,
+        touched,
+        isValid,
+        values,
+        submitForm,
+        validateForm,
+        isSubmitting,
+      }) => (
         <>
           <Form className="flex flex-col justify-center items-center md:justify-normal gap-2 w-full px-6 lg:w-1/2 max-w-xl flex-1">
             <div
@@ -208,11 +216,12 @@ export const DepositForm = () => {
                 }}
                 type="button"
                 className="flex-1 md:flex-[8]"
+                disabled={isSubmitting}
               >
                 {stepper.switch({
                   address: () => (isValid ? "review" : "next"),
                   amount: () => "next",
-                  confirm: () => "confirm",
+                  confirm: () => (isSubmitting ? "confirming..." : "confirm"),
                   status: () => "view history",
                 })}
               </FormButton>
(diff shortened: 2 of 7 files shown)


src/actions/hiro-fetch.ts
```diff
@@ -10,6 +10,7 @@ async function hiroFetch(url: string, options: RequestInit = {}) {
   return fetch(url, {
     ...options,
     headers: { ...options.headers, ...apiHeaders },
+    cache: "no-store",
```
djordon, line 13: What does this do? Does it instruct the Hiro API to not store the results?

src/app/reskin/withdraw/hooks/use-withdrawal-info.ts
```diff
@@ -27,5 +27,6 @@ export function useWithdrawalInfo({ txid }: { txid: string }) {
       }
       return 5000;
     },
+    enabled: !!txid,
```
djordon, line 30: Ahhh, this fixes the issue, since this only executes the query when the txid exists. So now it waits until it gets that from the wallet?

src/app/reskin/components/forms/deposit/deposit-form.tsx
```diff
@@ -143,7 +143,15 @@ export const DepositForm = () => {
...
     >
-      {({ errors, touched, isValid, values, submitForm, validateForm }) => (
+      {({
+        errors,
+        touched,
+        isValid,
+        values,
+        submitForm,
+        validateForm,
+        isSubmitting,
```
djordon, line 153: Does something need to be updated to supply the `isSubmitting` value?

djordon (commented):
Just had a few of questions.

src/actions/hiro-fetch.ts
hozzjss: it instructs next not to cache the results

djordon (dismissed):

src/app/reskin/components/stepper/deposit/status.tsx
```diff
@@ -40,7 +40,7 @@ export function TxPendingStatus({ bitcoinTxId }: { bitcoinTxId: string }) {
       <div>
         ( Estimation{" "}
         <HyperLink href={`${PUBLIC_MEMPOOL_URL}/tx/${bitcoinTxId}`}>
-          5min
+          1 hour
```
djordon, line 43: This is supposed to be 20 minutes for deposits.

hozzjss: it used to take signers some time to confirm it, I remember matteo adjusting private mainnet to be 3 blocks so even then it's 30 minutes

aldur: That's for withdrawals, here we are talking deposits

hozzjss: okay will change

hozzjss: fixed

aldur: @djordon it might simply be that @hozzjss was referring to a chance to withdrawals -- not to deposits.

djordon: I should've refreshed my page, I didn't need to write all of that.

djordon (approved):

aldur (approved):

PR #165, opened 2025-04-29
add validation for max decimal places in both deposits and withdrawals

hozzjss:
Resolves: #168

### DIFF
--- a/src/hooks/use-validate-deposit-amount.ts
+++ b/src/hooks/use-validate-deposit-amount.ts
@@ -24,7 +24,12 @@ export function useValidateDepositAmount({
       yup
         .number()
         .positive()
-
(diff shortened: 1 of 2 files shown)


src/hooks/use-validate-deposit-amount.ts
```diff
@@ -24,7 +24,12 @@ export function useValidateDepositAmount({
       yup
         .number()
         .positive()
-
+        .test("maxDigits", "BTC can have up to 8 decimal places", (number) => {
+          if (!number) return true;
+          const numberString = String(number);
+          const [, decimalPart] = numberString.split(".");
+          return decimalPart?.length <= 8 || decimalPart === undefined;
```
djordon, line 31: Does this automatically work with numbers like `0.0100000000000000`?

aldur (approved):

PR #166, opened 2025-04-29
add withdrawal max fee and withdrawal fee multiplier to health

hozzjss:


src/app/api/health/route.ts
```diff
@@ -9,6 +9,8 @@ export async function GET(_: NextRequest) {
         version: packageJson.version,
         contracts_deployer: env.SBTC_CONTRACT_DEPLOYER,
         git_commit: env.GIT_COMMIT || "unknown",
+        withdrawal_fee_multiplier: env.WITHDRAWAL_FEE_MULTIPLIER,
+        withdrawal_max_fee: 180 * env.WITHDRAWAL_FEE_MULTIPLIER,
```
aldur, line 13: Can we not have a magic number here and instead rely on a constant? Since this is also used elsewhere in the code.

aldur (changes requested):

src/app/api/health/route.ts
hozzjss: fair nuff

aldur (approved):

djordon (approved):
Looks good.

Side question: what is the reclaim fee and why is it fixed?

hozzjss:
> Looks good.
> 
> Side question: what is the reclaim fee and why is it fixed?
@djordon Vestiges of old Matteo and I thought about changing that to something similar to the withdrawal dynamic stuff but decided not to touch it since it would mean more testing when we were on a tight deadline, I think I might need to pick y'all's brains on that but I'd like to have it in a separate PR

PR #187, opened 2025-04-30
handle case where stx address is not available in deposits

hozzjss:


src/app/(reskin)/components/forms/deposit/deposit-form.tsx
```diff
@@ -203,16 +203,24 @@ export const DepositForm = () => {
                 buttonRef={nextButtonRef}
                 onClick={async () => {
                   const result = await validateForm();
```
matteojug, line 205: Another difference with the withdrawal one is that here we validate at the top, while there we validate the form after the early returns. If `validateForm` can throw, here we risk failing before reaching the early returns. Should we move `validateForm` right before `const error = result[currentStep];`?

matteojug (dismissed):
LGTM. Diffing this with the withdrawal one, here we are missing the `isConnected` check. Do we want/need that check here as well?

src/app/(reskin)/components/forms/deposit/deposit-form.tsx
hozzjss: it doesn't throw it returns an object with keys being the form field name and values being the error as a string but I can still make that run at the end, aaaand done!

matteojug (approved):

aldur (approved):

PR #209, opened 2025-05-21
update asigna to latest

hozzjss:
this should make asigna work with withdrawals like other wallets


matteojug (approved):
I didn't test this, only skimmed over the package diff

PR #212, opened 2025-05-29
feat:slider-with-balance-in-deposit-withraw-form

### LINKED ISSUE #181: [Feature]: add balance or "max" button to bridge reskin
# Feature - add balance or MAX button to bridge reskin

## 1. Description

I tried to do a simple thing many users often do -- withdraw all what I can. And it comes I need to do some mental arithmetics to understand how much should I put in "amount" field. Instead, it would be nice to have button "max"  (or even sliding bar allowing to choose x% from my balance).

ali-hs:
Add slider to deposit and withdraw forms. Max value of the slider represent btc/stbc balance of the connected wallet.

Resolves #181

https://github.com/user-attachments/assets/6978dfa1-dd37-4ebc-aa52-606f5e9f3b1f

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "@radix-ui/react-scroll-area": "1.2.4",
     "@radix-ui/react-select": "2.1.7",
     "@radix-ui/react-separator": "1.1.2",
+    "@radix-ui/react-slider": "^1.3.5",
     "@scure/bip32": "1.3.3",
     "@stacks/connect": "7.10.1",
     "@stacks/transactions": "7.0.5",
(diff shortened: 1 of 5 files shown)


aldur:
Thanks @ali-hs! Trying to simplify a bit UX, what if we instead just add a "max" button?

ali-hs:
> Thanks @ali-hs! Trying to simplify a bit UX, what if we instead just add a "max" button?

@aldur Both in the deposit and withdraw form? 

Should this button replace the slider in the above UI with the following titles?
Deposit Form:
        "Deposit All <max_btc_amount>"
 
 Withdraw Form:
        "Withdraw All <max_sbtc_amount>"

aldur:
I think that looks correct, yes please!

MCJOHN974:
btw I don't think max button is better then slider, I sometimes want to convert "~around X%" of this wallet balance from one currency to another, and in this cases slider is really a great ui

PR #217, opened 2025-07-16
feat:bns-support-in-deposit-flow

ali-hs:
This PR add support for BNS in deposit flow. It only supports bns-v2

https://github.com/user-attachments/assets/28d6ecff-741c-417c-bc2f-66f1de0e9a20


package.json
```diff
@@ -26,6 +26,7 @@
     "@tanstack/react-query": "5.61.5",
     "bitcoin-address-validation": "3.0.0",
     "bitcoinjs-lib": "7.0.0-rc.0",
+    "bns-v2-sdk": "^2.0.0",
```
matteojug, line 29: We usually pin the exact version we are using

src/actions/bns-client.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export async function addressToBnsName(
+  address: string,
+): Promise<Array<{ name: string; namespace: string }>> {
+  const network: NetworkType = getNetwork();
+  return fetchUserOwnedNames({
+    senderAddress: address,
+    network: network,
+  });
+}
```
matteojug, line 24: nit: this one doesn't appear to be used around, do we need it?

src/actions/bns-client.ts
```diff
@@ -0,0 +1,28 @@
...
+import { env } from "@/env";
+
+export type NetworkType = "mainnet" | "testnet";
+
+export async function bnsNameToAddress(name: string): Promise<string | null> {
+  const network: NetworkType = getNetwork();
+  return getOwner({
+    fullyQualifiedName: name,
+    network,
+  });
```
matteojug, line 13: This seems to be using `api.bnsv2.com` (before falling back to the read only contract call). I'm wondering if instead we should replace this with a readonly contract call, to avoid the trust assumption.

src/util/bns.ts
```diff
@@ -0,0 +1,7 @@
+export function isValidBNSName(name: string): boolean {
+  if (name) {
+    const dotIndex = name.indexOf(".");
+    return dotIndex > 0 && dotIndex < name.length - 1; // dot comes in between
```
matteojug, line 4: I'm not 100% sure about the BNS format, but `"a.b."` would pass this filter, and I don't think it's a valid name. Looking at BNS hub, it says _Only lowercase letters, numbers, hyphens, and underscores are allowed_ when searching for a domain, and the smart contract seem to have a similar logic. But then in BNS hub I see emojii, so maybe there's something else going on.

Anyway, IIUC this is just to avoid "wasting" queries to BNS, so it's not an issue if we have false negatives. But we could improve this a little bit.

ali-hs:
@matteojug I have address the feedback. Can you please have a look again?

PR #230, opened 2025-10-31
[feat]: cache Emily limits

MCJOHN974:
Currently, the bridge polls Emily limits quite frequently. However, we do not change limits often (usually once per few months or so). Thus, this caching can eliminate a lot of Emily polls, without harming user experience too much.

Note: I choosed caching time to 1 bitcoin block more or less out of thin air and do not have a strong opinion about if this should be this amount of time or another.

### DIFF
--- a/src/actions/get-emily-limits.ts
+++ b/src/actions/get-emily-limits.ts
@@ -1,18 +1,20 @@
 "use server";
 
 import { env } from "@/env";
+
 type EmilyLimits = {
   pegCap: null | number;
   perDepositCap: null | number;
   perWithdrawalCap: null | number;
-  accountCaps: AccountCaps;
+  accountCaps: Record<string, unknown>;
   perDepositMinimum: null | number;
   availableToWithdraw: number | null;
 };
 
-type AccountCaps = {};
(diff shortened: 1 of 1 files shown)


src/actions/get-emily-limits.ts
```diff
@@ -33,3 +37,19 @@ export default async function getEmilyLimits() {
     availableToWithdraw: json.availableToWithdraw ?? Infinity,
   };
 }
+
+export default async function getEmilyLimits() {
+  const now = Date.now();
+  const TEN_MINUTES = 10 * 60 * 1000;
```
djordon: We'll probably want to shorten this to every 15 seconds or something. We want to be sure that if we shut off the bridge by changing the limits, users will not be able to use it.

djordon (commented):
I haven't looked too deeply into this, just trying to think about the caching policy. We'll probably want to choose something like 15 seconds or so to make sure that a change in limits is promptly reflected on the bridge.

src/actions/get-emily-limits.ts
fabrizio-stacks: Can't we use the api gw cache?

djordon:
We're doing work on the bridge in a new repo, so closing this.

PR #245, opened 2026-01-15
feat: show BTC and sBTC balances

heidar-stacks:
- Show BTC and sBTC balances besides input field
- Add "MAX" functionality to input max token amount
- Configure Vitest to run tests on CI
- Add unit test


.github/workflows/ci.yml
```diff
@@ -25,8 +25,14 @@ jobs:
...
           node-version: ${{ matrix.node-version }}
 
-      - name: Yarn Build
+      - name: Install Dependencies
+        id: yarn_install
+        run: yarn install --frozen-lockfile
+
+      - name: Run Tests
+        id: yarn_test
+        run: yarn test
```
matteojug, line 34: qq: why test before build, and not the other way around?

src/hooks/use-sbtc-balance.ts
```diff
@@ -1,29 +1,41 @@
...
-    // actions are mutations that are not expected to run
-    // at initial render that's why this workaround exists
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [address]);
+    const fetchBalance = async () => {
+      setIsLoading(true);
 
-  return query;
+      try {
+        const url = `${STACKS_API_URL}/extended/v2/addresses/${address}/balances/ft`;
```
matteojug, line 20: Doing this client side we cannot use an API key (previously set via `env.HIRO_API_KEY`), so we risk getting rate limited

matteojug, line 20: Is there a reason to prefer the `balances/ft` vs the previous read only call? Also, don't we need to handle pagination if done this way?

djordon, line 20: Is there a way to ask the wallet for the sBTC balance?

src/hooks/use-btc-balance.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { MEMPOOL_API_URL } = useAtomValue(bridgeConfigAtom);
+
+  useEffect(() => {
+    if (!address || !MEMPOOL_API_URL) return;
+
+    const fetchBalance = async () => {
+      setIsLoading(true);
+
+      try {
+        const url = `${MEMPOOL_API_URL}/address/${address}`;
```
djordon, line 19: We need to be careful here, since an "account" can have more than one BTC address associated with it. This can happen, for example, when a Leather wallet is connected to a Ledger device. By default Ledger will generate a unique address for receiving funds for all accounts created in the wallet. This means that it's possible that some of the user's coins are controlled by different addresses. If possible, we should ask the wallet for their BTC balance and use that here.

We already spoke about this offline, just posting here for visibility.

PR #250, opened 2026-02-06
feat: missing publickey error handling

heidar-stacks:
Show error with info when wallet key is unavailable during deposit


matteojug (approved):
LGTM, can you also add it to for the reclaim flow? We currently just error with `Please connect your wallet`

PR #258, opened 2026-04-20
fix: remove withdrawals paragraphs from ToC

adriano-stacks:
Use an online PDF diff tool to check for the delta there.


matteojug-stacks (approved):
PDF diff seems legit

djordon-stacks (approved):
Looks good

